PR #169, opened 2026-04-20
fix: protobuf parser resyncs past leading garbage instead of stalling

tylerkron:
## Summary
- Removes the `MaxRetryAttempts = 3` cap in `ProtobufMessageParser`. After 3 failed parse attempts the loop exited without advancing past unrecognizable bytes — the consumer buffer stayed clogged forever and downstream `MessageReceived` events stopped firing.
- Replaces the cap with single-byte resync on every failure (malformed varint, oversized length, parse exception). The parser now drains until either a real frame is recovered or the buffer is empty.
- Adds regression test `ProtobufMessageParser_ParseMessages_RecoversFromLeadingGarbage` that prepends 20 zero bytes to a valid `WriteDelimitedTo` frame and asserts the parser recovers `MsgTimeStamp = 42` and consumes the full input.

## Why this matters
This regression manifests as the desktop Live Graph receiving zero stream messages after `StartStreaming`. The DTR pulse on Microchip USB-to-Serial chips emits a few boot-time bytes when the port opens; those land at the head of the consumer buffer. Pre-fix, `QueuedMessageCount` (which is byte count, not message count) climbed from 66 → 8202 in 2 seconds while no frames were dispatched. Post-fix, the parser advances past the boot garbage and the first valid streaming frame is recovered.

## Test plan
- [x] `dotnet test --filter "FullyQualifiedName~ProtobufMessageParser" -f net9.0` — 8/8 pass (new test green)
- [x] `dotnet test --filter "FullyQualifiedName~Consumers" -f net9.0` — 41/41 pass (no regressions)
- [ ] Smoke test with desktop streaming flow against `/dev/cu.usbmodem101` after the new Core release lands

### DIFF
--- a/src/Daqifi.Core.Tests/Communication/Consumers/ProtobufMessageParserTests.cs
+++ b/src/Daqifi.Core.Tests/Communication/Consumers/ProtobufMessageParserTests.cs
@@ -105,6 +105,45 @@ public class ProtobufMessageParserTests
     }
 
     [Fact]
+    public void ProtobufMessageParser_ParseMessages_RecoversFromLeadingGarbage()
+    {
+        // Reproduces the real-world failure where boot-time garbage on the serial port
+        // (DTR pulse on Microchip USB-to-Serial chips, partial frames, etc.) leaves
+        // unrecognizable bytes at the head of the consumer buffer. When a valid streaming
+        // frame eventually arrives, the parser must skip past the garbage and parse the
+        // valid frame instead of stalling forever.
+        //
+        // With the original 3-retry cap, the parser exits after 3 failed attempts and
+        // never advances past the garbage; subsequent reads keep stacking bytes onto a
+        // buffer the parser refuses to drain.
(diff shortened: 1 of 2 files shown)


tylerkron:
**Re: Qodo "Resync allocates per attempt" — addressed in d9aa65b**

Fixed the allocation by switching to `MessageParser<T>.ParseFrom(ReadOnlySpan<byte>)` (available in Google.Protobuf 3.18+, we're on 3.34.1). Eliminates the per-attempt `byte[]` copy entirely while keeping the slice/length math the same.

Not addressing the second part of Qodo's note (exception-as-control-flow) in this PR — short of reimplementing protobuf parsing, the only way to know whether a candidate length-prefixed slice is valid is to attempt to parse it. The garbage-recovery path is already a cold/exceptional case (real streaming traffic parses the first try and never hits the catch); the allocation fix removes the dominant cost.

PR #175, opened 2026-04-25
chore: upgrade target frameworks from net8/net9 to net9/net10

tylerkron:
## Summary

- Drops .NET 8 support, adds .NET 10 support to always track the latest two stable releases
- Updates `TargetFrameworks` in both `Daqifi.Core.csproj` and `Daqifi.Core.Tests.csproj` from `net8.0;net9.0` to `net9.0;net10.0`
- Updates `dotnet-version` in both `ci.yml` and `release.yml` from `8.0.x / 9.0.x` to `9.0.x / 10.0.x`

## Test plan

- [ ] CI build passes for both `net9.0` and `net10.0` targets
- [ ] Tests pass on both frameworks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,8 @@ jobs:
       uses: actions/setup-dotnet@v5
       with:
         dotnet-version: |
-          8.0.x
           9.0.x
+          10.0.x
         
     - name: Restore dependencies
       run: dotnet restore Daqifi.Core.sln
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,8 +19,8 @@ jobs:
       uses: actions/setup-dotnet@v5
       with:
         dotnet-version: |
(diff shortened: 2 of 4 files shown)


tylerkron:
Re: Qodo issue — **Release SDKs not validated**: Disagree — no change needed.

When a project multi-targets `net9.0;net10.0`, a single `dotnet build` / `dotnet test` invocation with the .NET 10 SDK builds and tests *both* TFMs in one pass. A matrix is not needed.

The reason both `9.0.x` and `10.0.x` are listed in `setup-dotnet` is that `dotnet test` needs the **.NET 9 runtime** present to execute the `net9.0` test binaries — the .NET 10 SDK ships with the .NET 10 runtime only. Installing `9.0.x` makes the .NET 9 runtime available so tests for both TFMs actually run.

Switching to only `10.0.x` (Qodo's Option B) would silently skip the `net9.0` test run, which is worse. No change needed.

PR #180, opened 2026-04-28
fix: bind UDP discovery per-NIC and skip virtual adapters (#179)

### LINKED ISSUE #179: UDP discovery fails on hosts with virtual NICs on same subnet (WSL2 mirrored, Hyper-V, etc.) — bind broadcast per-NIC + filter virtual adapters
## Problem

`WiFiDeviceFinder` enumerates per-NIC and sends a subnet-directed broadcast for each, but uses a single `UdpClient` bound to `IPAddress.Any` for both sending and receiving. On hosts where a virtual NIC (WSL2 mirrored networking, Hyper-V vEthernet, VPN/TAP tunnels) shares the same /24 subnet as the real WiFi/Ethernet adapter, discovery silently fails:

- Windows routing table picks the wrong egress NIC for each \`Send\`
- Replies that come back on the real NIC may be tagged as \"from the virtual NIC\" via subnet-match logic
- Result: discovery returns zero devices even though the device is reachable via TCP from the same host

## Reproduction

Windows host with WSL2 mirrored networking enabled. WSL2 mirrored mode creates a Windows-visible \"Ethernet 3\" / \"vEthernet (WSL)\" adapter on the same subnet as the physical WiFi (e.g. WiFi at 192.168.1.100, vEthernet at 192.168.1.133, both on 192.168.1.0/24).

PowerShell verification:

\`\`\`powershell
# Bind explicitly to physical WiFi → BOTH devices reply
\$client = New-Object System.Net.Sockets.UdpClient(0)
\$client.EnableBroadcast = \$true
\$client.Client.Bind([IPEndPoint]::new([IPAddress]::Parse(\"192.168.1.100\"), 0))
\$client.Send([Encoding]::ASCII.GetBytes(\"discover\"), 8, [IPEndPoint]::new([IPAddress]::Parse(\"255.255.255.255\"), 30303))
# → REPLY 120 bytes from 192.168.1.60:30303
# → REPLY 120 bytes from 192.168.1.160:30303

# Default bind 0.0.0.0 (what Core does) → ZERO replies
\$client = New-Object System.Net.Sockets.UdpClient(0)
\$client.EnableBroadcast = \$true
# (no Bind call, defaults to 0.0.0.0)
\$client.Send(…)
# → no replies
\`\`\`

Same network, same devices, same broadcast destination — only difference is whether the client socket is bound to a specific NIC IP or left at \`IPAddress.Any\`.

Discovery works on a different PC (no WSL) on the same Tesla AP — the problem is host-specific and reproduces consistently in our WSL2 test environment.

## Code locations

\`src/Daqifi.Core/Device/Discovery/WiFiDeviceFinder.cs\`
- Line 147: \`using var udpTransport = new UdpTransport(_discoveryPort)\` — single socket
- Line 168-169 loop: per-NIC \`SendBroadcastAsync\` from same socket → Windows picks egress NIC, not reliably the per-NIC iteration's NIC
- Line 303-311 \`GetAllNetworkInterfaces\` filter: \`OperationalStatus.Up\` + IPv4 + (Ethernet || Wireless80211) — passes WSL vEthernet / Hyper-V vEthernet / TAP tunnels
- Line 362-403 \`FindLocalInterfaceForRemoteAddress\`: subnet-match returns first NIC whose \`(local & mask) == (remote & mask)\` — order-dependent on virtual NICs sharing subnet

\`src/Daqifi.Core/Communication/Transport/UdpTransport.cs\`
- Line 94 \`new UdpClient(_localPort)\` binds to \`IPAddress.Any\`. No constructor overload for binding to a specific local IP.

## Suggested fix

Three changes to \`WiFiDeviceFinder\`:

### 1. Bind a separate `UdpClient` per NIC for the broadcast send

\`\`\`csharp
foreach (var nic in networkInterfaces)
{
    using var udp = new UdpClient(new IPEndPoint(nic.LocalAddress, 0));
    udp.EnableBroadcast = true;
    await udp.SendAsync(_queryCommandBytes, _queryCommandBytes.Length, nic.BroadcastEndpoint);
}
\`\`\`

This guarantees the broadcast leaves via the intended NIC and the source IP is stamped correctly so replies come back to the correct adapter.

### 2. Add virtual-adapter name/description filter

\`\`\`csharp
var name = networkInterface.Name ?? \"\";
var desc = networkInterface.Description ?? \"\";
if (name.StartsWith(\"vEthernet\", StringComparison.OrdinalIgnoreCase) ||
    desc.IndexOf(\"vEthernet\", StringComparison.OrdinalIgnoreCase) >= 0 ||
    desc.IndexOf(\"Hyper-V\", StringComparison.OrdinalIgnoreCase) >= 0 ||
    desc.IndexOf(\"WSL\", StringComparison.OrdinalIgnoreCase) >= 0 ||
    desc.IndexOf(\"VirtualBox\", StringComparison.OrdinalIgnoreCase) >= 0 ||
    desc.IndexOf(\"VMware\", StringComparison.OrdinalIgnoreCase) >= 0 ||
    desc.IndexOf(\"TAP\", StringComparison.OrdinalIgnoreCase) >= 0)
{
    continue;
}
\`\`\`

### 3. Use the actual receive socket's NIC for `LocalInterfaceAddress`

Per-NIC sockets means each socket knows its own bound IP. When a reply arrives, the socket that received it IS the local interface — no subnet-match guess needed. Populate \`LocalInterfaceAddress\` from the receiving \`UdpClient\`'s local endpoint.

## Risks / considerations

- Per-NIC \`UdpClient\` instances increase socket count for the discovery duration. Negligible on any modern host — typically 1-3 NICs after virtual-adapter filtering.
- Some users may rely on virtual adapters being eligible (e.g. development against a virtualized device). Could expose the virtual-adapter filter via opt-in / configuration.
- \`Wireless80211\` filter matters for WiFi-only setups. Confirm Ethernet-attached devices still discover (unlikely a regression — code path is symmetric).

## Acceptance criteria

- Discovery succeeds on Windows hosts with WSL2 mirrored networking (vEthernet on same /24)
- Discovery succeeds on Hyper-V hosts with virtual switches on host NIC subnet
- \`LocalInterfaceAddress\` reflects the NIC that actually received the reply, not a subnet-match guess
- Existing single-NIC scenarios continue to work (regression check)
- Unit test or integration test with mocked NIC list including a virtual adapter

## Related tickets (not duplicates)

- daqifi-core #147 — Honor LocalInterfaceAddress when **connecting** discovered WiFi devices (TCP stage; this ticket is the UDP discovery stage)
- daqifi-desktop #352 — Bind TCP connections to correct local interface for multi-NIC support (TCP stage; same)
- daqifi-core #88 (closed) — Add LocalInterfaceAddress tracking; closed but the implementation is incomplete in the WSL-vEth-on-same-subnet edge case

cptkoolbeenz:
Closes #179.

## Summary
- `WiFiDeviceFinder` now binds **one `UdpClient` per NIC** at `(LocalAddress, 0)` for both send and receive. The OS routes each broadcast out the intended adapter; replies arrive on the same socket; the socket's bound IP becomes the authoritative `LocalInterfaceAddress` (no more subnet-match guessing across virtual NICs that share a subnet).
- `GetAllNetworkInterfaces` now skips virtual/tunnel adapters by Name/Description (`vEthernet`, `Hyper-V`, `WSL`, `VirtualBox`, `VMware`, `TAP`).
- Removed the now-dead `FindLocalInterfaceForRemoteAddress` helper and unused `SubnetMask` field on `NetworkInterfaceInfo`.
- Added `IsVirtualOrTunnelInterface` (`internal static`, `InternalsVisibleTo` already in place) plus xUnit `Theory` cases covering common virtual-adapter shapes and physical-adapter negatives.

## Why
On Windows hosts with WSL2 mirrored networking (or Hyper-V/VPN/TAP) the virtual NIC frequently shares a `/24` with the real WiFi/Ethernet adapter. Sending broadcasts from a single `UdpClient` bound to `IPAddress.Any` lets Windows pick the wrong egress NIC, and the subnet-match reply heuristic could attribute a reply to the virtual NIC. Discovery would silently return zero devices even though the device was reachable via TCP from the same host.

## Behavior note
`DeviceDiscovered` may now be invoked **concurrently** from multiple per-NIC receive loops (was strictly sequential before). The existing public contract did not guarantee single-threaded invocation; subscribers should already be thread-safe.

## Test plan
- [x] `dotnet build Daqifi.Core.sln` — clean (net8.0 + net9.0), 0 warnings
- [x] `dotnet test Daqifi.Core.sln` — 830 passed / 0 failed / 2 skipped (both target frameworks)
- [x] New tests `IsVirtualOrTunnelInterface_*` cover vEthernet (WSL), vEthernet (Default Switch), `Ethernet 3` w/ Hyper-V description, `Ethernet 4` w/ WSL description, VirtualBox, VMware, TAP-Windows — plus Wi-Fi/Ethernet/WLAN negatives and null inputs
- [x] Manually verified on a Windows host (Hyper-V `vEthernet (Default Switch)` 172.25.176.1 + Intel 10G `Ethernet 3` 192.168.1.133): DAQiFi-95A7 at 192.168.1.160 was discovered and `LocalInterfaceAddress` correctly populated as `192.168.1.133` (the real NIC, not the Hyper-V adapter)
- [ ] Strict WSL2-mirrored-on-same-subnet repro is best-validated by issue reporter

### DIFF
--- a/src/Daqifi.Core.Tests/Device/Discovery/WiFiDeviceFinderTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Discovery/WiFiDeviceFinderTests.cs
@@ -114,6 +114,10 @@ public class WiFiDeviceFinderTests
     [InlineData("VirtualBox Host-Only Network", "VirtualBox Host-Only Ethernet Adapter")]
     [InlineData("VMware Network Adapter VMnet1", "VMware Virtual Ethernet Adapter for VMnet1")]
     [InlineData("OpenVPN TAP-Windows6", "TAP-Windows Adapter V9")]
+    // Keyword present only in name (not description) — both fields must be checked.
+    [InlineData("vEthernet (Default Switch)", "")]
+    [InlineData("WSL Bridge", "Generic Adapter")]
+    [InlineData("VMware NAT", "Generic Adapter")]
     public void IsVirtualOrTunnelInterface_VirtualAdapters_ReturnsTrue(string name, string description)
     {
         Assert.True(WiFiDeviceFinder.IsVirtualOrTunnelInterface(name, description));
--- a/src/Daqifi.Core.Tests/Device/Discovery/WiFiDeviceFinderTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Discovery/WiFiDeviceFinderTests.cs
@@ -99,4 +99,37 @@ public class WiFiDeviceFinderTests
         // Assert
         Assert.NotNull(finder);
     }
+
+    // Virtual / tunnel adapter filter — issue #179.
+    // WSL2 mirrored networking, Hyper-V vEthernet, VPN/TAP adapters frequently share a /24
+    // subnet with the real WiFi/Ethernet NIC and cause Windows routing to pick the wrong egress
+    // for broadcasts, silently breaking discovery.
+    [Theory]
+    [InlineData("vEthernet (WSL)", "Hyper-V Virtual Ethernet Adapter")]
+    [InlineData("vEthernet (Default Switch)", "Hyper-V Virtual Ethernet Adapter")]
+    [InlineData("Ethernet 3", "Hyper-V Virtual Ethernet Adapter #2")]
+    [InlineData("Ethernet 4", "WSL Virtual Ethernet Adapter")]
+    [InlineData("VirtualBox Host-Only Network", "VirtualBox Host-Only Ethernet Adapter")]
+    [InlineData("VMware Network Adapter VMnet1", "VMware Virtual Ethernet Adapter for VMnet1")]
+    [InlineData("OpenVPN TAP-Windows6", "TAP-Windows Adapter V9")]
+    public void IsVirtualOrTunnelInterface_VirtualAdapters_ReturnsTrue(string name, string description)
+    {
+        Assert.True(WiFiDeviceFinder.IsVirtualOrTunnelInterface(name, description));
+    }
+
+    [Theory]
+    [InlineData("Wi-Fi", "Intel(R) Wi-Fi 6 AX201 160MHz")]
+    [InlineData("Ethernet", "Realtek PCIe GbE Family Controller")]
+    [InlineData("Ethernet 2", "Intel(R) Ethernet Connection I219-LM")]
+    [InlineData("WLAN", "Qualcomm Atheros QCA9377 Wireless Network Adapter")]
+    public void IsVirtualOrTunnelInterface_PhysicalAdapters_ReturnsFalse(string name, string description)
+    {
+        Assert.False(WiFiDeviceFinder.IsVirtualOrTunnelInterface(name, description));
+    }
+
+    [Fact]
+    public void IsVirtualOrTunnelInterface_NullInputs_ReturnsFalse()
+    {
+        Assert.False(WiFiDeviceFinder.IsVirtualOrTunnelInterface(null, null));
+    }
 }
--- a/src/Daqifi.Core.Tests/Device/Discovery/WiFiDeviceFinderTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Discovery/WiFiDeviceFinderTests.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
+using System.Net.NetworkInformation;
 using System.Threading;
 using System.Threading.Tasks;
 using Daqifi.Core.Device.Discovery;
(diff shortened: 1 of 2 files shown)


cptkoolbeenz:
Pass 1 skip notes:

- **Avoid indefinite discovery hang** (importance 1/10) — SKIPPED. Qodo's own "Why" explicitly states the change "would alter the API's intended semantics". The infinite-timeout overload is part of the public contract and matches pre-fix behavior; callers who want a bounded scan use `DiscoverAsync(TimeSpan)`.

PR #182, opened 2026-05-02
feat: distinguish SD card error states in GetSdCardFilesAsync

### LINKED ISSUE #181: Distinguish SD card error states in GetSdCardFilesAsync (don't swallow SCPI errors)
## Problem

`DaqifiStreamingDevice.GetSdCardFilesAsync` currently silently swallows SCPI errors and returns an empty list, which makes any failure in the device-side LIST path indistinguishable from "the SD card is empty." We hit this debugging a real-world case where a corrupt SD card caused `SD:LISt?` to reliably return `**ERROR -200, "Execution error"` on a perfectly good firmware (v3.4.4). Both the Core CLI's `--sd-list` and Daqifi.Desktop's Device Logs tab just showed "no files" with zero indication anything was wrong, sending us on a multi-hour rabbit hole through the firmware before someone thought to swap the card.

Today's flow in [`DaqifiStreamingDevice.cs:294-358`](src/Daqifi.Core/Device/DaqifiStreamingDevice.cs):

1. Send `SYSTem:STORage:SD:LISt?`
2. If response contains a SCPI error → retry once
3. If still error → fall through to `SdCardFileListParser.ParseFileList(lines)`
4. The parser drops every line starting with `**ERROR` ([SdCardFileListParser.cs:44-48](src/Daqifi.Core/Device/SdCard/SdCardFileListParser.cs))
5. Return empty `IReadOnlyList<SdCardFileInfo>`

A consumer cannot distinguish:
- ✅ Card is present, FS is fine, directory is empty
- ❌ No SD card installed
- ❌ Card present but FS corrupt / mount failing / directory unreadable
- ❌ Card busy (logging in progress)
- ❌ SD subsystem not enabled

## What the firmware actually tells us

The firmware emits distinguishable text alongside the SCPI error code (see [SCPIStorageSD.c:77-97](https://github.com/daqifi/daqifi-nyquist-firmware/blob/main/firmware/src/services/SCPI/SCPIStorageSD.c#L77-L97) and [`SCPI_StorageSDListDir`](https://github.com/daqifi/daqifi-nyquist-firmware/blob/main/firmware/src/services/SCPI/SCPIStorageSD.c#L227-L287)):

| Condition | Wire response |
| --- | --- |
| No SD card present | `\r\nError !! No SD Card Detected\r\n` + `**ERROR: -200, "Execution error"` |
| SD not enabled (in Get/Logging paths) | `\r\nError !! Please Enabled SD Card\r\n` + `**ERROR: -200` |
| Directory open failed (FS issue) | `\r\n[Error:N]Failed to open directory [path]\r\n` |
| Card busy | bare `**ERROR: -200, "Execution error"` |
| Operation timeout | bare `**ERROR: -200, "Execution error"` |

We're already collecting these lines — we're just throwing the diagnostic information away.

## Proposed change

Define a small result/exception model and let `GetSdCardFilesAsync` surface what actually happened:

```csharp
public class SdCardNotPresentException : SdCardOperationException { }
public class SdCardBusyException        : SdCardOperationException { }
public class SdCardFilesystemException  : SdCardOperationException { public string DeviceMessage { get; } }
public class SdCardOperationException   : Exception {
    public IReadOnlyList<string> RawDeviceResponse { get; }
    public string? LastScpiError { get; }
}
```

After the retry loop, before parsing, classify the response:

- Contains `"No SD Card Detected"` (case-insensitive) → throw `SdCardNotPresentException`
- Contains `"Failed to open directory"` → throw `SdCardFilesystemException` with the `[Error:N]...` text
- Contains `**ERROR` lines but no file lines → throw `SdCardOperationException` with the SCPI error code/message
- Contains file lines (whether or not error lines are interleaved) → return the parsed list (current behavior)

The successful "device returned 0 files" case stays untouched.

## UX impact

This unblocks the desktop UX fix tracked in daqifi/daqifi-desktop ([link to be added once filed]). With these exceptions, the Device Logs tab can show actionable messages like "No SD card installed" or "SD card error — check that the card is formatted FAT32" instead of the current "NO FILES — Files will appear here when the device logs data to its SD card."

## Repro

Original session that uncovered this: corrupt SD card on a `Nq1` running firmware `3.4.4`. `--sd-log-start --duration 5` succeeded, `--sd-list` returned 0 files, raw serial confirmed the firmware was returning `**ERROR: -200, "Execution error"` on every `SD:LISt?` call. Swapping in a fresh card made it work immediately.

tylerkron:
## Summary

Closes #181.

`GetSdCardFilesAsync` silently swallowed SCPI errors and returned an empty list, making real failures (no card, corrupt FS, busy SD subsystem) indistinguishable from \"directory is empty.\" A corrupt SD card on firmware 3.4.4 sent multiple debug sessions chasing phantom firmware bugs before someone thought to swap the card.

This adds a typed exception hierarchy and classifies the response after the existing retry loop so callers can show actionable detail.

### What changed

- **New exception hierarchy** in `Daqifi.Core.Device.SdCard`:
  - `SdCardOperationException` — base; carries `RawDeviceResponse` and `LastScpiError`
  - `SdCardNotPresentException` — \"No SD Card Detected\"
  - `SdCardFilesystemException` — \"Failed to open directory\" with the device's raw error text in `DeviceMessage`
  - `SdCardBusyException` — defined for completeness (firmware cannot distinguish busy from timeout on the wire, so this currently surfaces as the base type)
- **`DaqifiStreamingDevice.GetSdCardFilesAsync`** now calls a private `ThrowIfSdCardListError` classifier after the retry loop. Specific firmware markers take precedence over the generic content/error check; an empty directory still returns an empty list.
- **`ISdCardOperations`** xmldoc updated to document the new exceptions.

### Behavior preserved

- Empty directory (no errors, no content lines) → returns empty list (unchanged).
- File lines present, even with stray interleaved `**ERROR` lines → returns parsed files (unchanged).
- LAN interface is still restored in the `finally` block on the throwing path.

### Hardware verification

Tested on `/dev/cu.usbmodem2101` via the example app:
- Card with files → returns files (no exception)
- No card → throws `SdCardNotPresentException` (previously silently returned 0 files)
- A \"corrupt\" card available for testing returned an empty `SD:LISt?` response at the firmware level (i.e. firmware mounted it and saw 0 files), so the `SdCardFilesystemException` and bare-`**ERROR` paths are exercised by unit tests against the firmware wire formats from the issue.

## Test plan

- [x] Unit tests pass on net9.0 + net10.0 (878 passing, 0 failing)
- [x] New tests cover: not-present, filesystem-error (with and without SCPI co-error), persistent-SCPI-error, files-with-interleaved-error, empty-directory, LAN-restore-on-error
- [x] Updated test that previously asserted empty list on persistent SCPI error now asserts `SdCardOperationException`
- [x] Manual verification on physical device for happy path and not-present path

### DIFF
--- a/src/Daqifi.Core.Tests/Device/SdCard/SdCardOperationsTests.cs
+++ b/src/Daqifi.Core.Tests/Device/SdCard/SdCardOperationsTests.cs
@@ -681,6 +681,54 @@ namespace Daqifi.Core.Tests.Device.SdCard
         }
 
         [Fact]
+        public async Task GetSdCardFilesAsync_LastScpiError_ContainsOnlyScpiFormattedLine()
+        {
+            // Arrange — firmware emits both "Error !! ..." status text AND a SCPI error.
+            // LastScpiError must only carry the SCPI-formatted line so callers can
+            // parse it; the status text is preserved in RawDeviceResponse.
+            var device = new RetryableSdCardStreamingDevice("TestDevice");
+            var response = new List<string>
+            {
+                "Error !! No SD Card Detected",
+                "**ERROR: -200, \"Execution error\""
+            };
+            device.ResponseSequence.Enqueue(response);
+            device.ResponseSequence.Enqueue(new List<string>(response));
+            device.Connect();
+
+            // Act
+            var ex = await Assert.ThrowsAsync<SdCardNotPresentException>(
+                () => device.GetSdCardFilesAsync());
+
+            // Assert — LastScpiError must be the SCPI line, never the firmware text
+            Assert.NotNull(ex.LastScpiError);
+            Assert.StartsWith("**ERROR", ex.LastScpiError);
+            Assert.DoesNotContain("Error !!", ex.LastScpiError);
+            Assert.Contains("Error !! No SD Card Detected", ex.RawDeviceResponse);
+        }
+
+        [Fact]
+        public async Task GetSdCardFilesAsync_WithFirmwareTextOnly_ThrowsWithNullLastScpiError()
+        {
+            // Arrange — defensive: hypothetical firmware response with status text
+            // but no SCPI error line. Shouldn't happen for known paths, but the
+            // classifier must not silently return an empty list.
+            var device = new RetryableSdCardStreamingDevice("TestDevice");
+            var response = new List<string> { "Error !! Some unfamiliar firmware error" };
+            device.ResponseSequence.Enqueue(response);
+            device.ResponseSequence.Enqueue(new List<string>(response));
+            device.Connect();
+
+            // Act
+            var ex = await Assert.ThrowsAsync<SdCardOperationException>(
+                () => device.GetSdCardFilesAsync());
+
+            // Assert
+            Assert.Null(ex.LastScpiError);
+            Assert.Contains("Some unfamiliar firmware error", ex.Message);
+        }
+
+        [Fact]
         public async Task GetSdCardFilesAsync_OnError_StillRestoresLanInterface()
         {
             // Arrange - persistent error path must still restore the LAN interface
--- a/src/Daqifi.Core.Tests/Device/SdCard/SdCardOperationsTests.cs
+++ b/src/Daqifi.Core.Tests/Device/SdCard/SdCardOperationsTests.cs
@@ -552,20 +552,151 @@ namespace Daqifi.Core.Tests.Device.SdCard
         }
 
         [Fact]
-        public async Task GetSdCardFilesAsync_WithPersistentScpiError_ReturnsEmptyAfterRetries()
+        public async Task GetSdCardFilesAsync_WithPersistentScpiError_ThrowsSdCardOperationException()
         {
-            // Arrange - simulate persistent error
+            // Arrange - simulate persistent bare SCPI error (card busy / timeout territory).
+            // Previous behavior returned an empty list, which made real failures look
+            // identical to "directory is empty". Issue #181 surfaces this as a typed
+            // exception so callers can show actionable detail.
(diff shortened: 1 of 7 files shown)


tylerkron:
Thanks @qodo-code-review — both findings addressed in d0dda48.

**Bug #1 — `LastScpiError` misclassified**: Agreed. Tightened `IsScpiErrorLine` to require `**ERROR` or `ERROR:` (with the colon), so firmware status text like `Error !! No SD Card Detected` no longer ends up in `LastScpiError`. The non-SCPI text is still preserved in `RawDeviceResponse` for diagnostics. Added a separate `IsNonResultLine` (permissive) for the `hasContentLine` check, so firmware text is still recognized as a non-result line — just not as a SCPI error.

Also added a defensive fallback in `ThrowIfSdCardListError`: if a response contains only firmware status text with no SCPI error and no recognized marker, we throw `SdCardOperationException` (with `LastScpiError = null`) rather than silently returning an empty list. Shouldn't happen for known firmware paths, but it's a meaningful safety net for future firmware revisions.

**Bug #2 — Duplicate predicate**: Agreed. `ContainsScpiError` now delegates to `IsScpiErrorLine` so retry triggering and classification share a single source of truth.

New tests cover both behaviors:
- `GetSdCardFilesAsync_LastScpiError_ContainsOnlyScpiFormattedLine` — verifies `LastScpiError` is the SCPI line, not the firmware text, even when both are present.
- `GetSdCardFilesAsync_WithFirmwareTextOnly_ThrowsWithNullLastScpiError` — verifies the defensive fallback path.

Full suite still green: 880 passing on net9.0 + net10.0.

PR #185, opened 2026-05-05
feat: add DrainErrorQueueAsync for SCPI error queue inspection

### LINKED ISSUE #174: feat(scpi): add SYST:ERR? drain helper returning popped errors for logging/diagnostics
## Summary

Daqifi.Core's SCPI error story today is pure inline scanning: \`ContainsScpiError(IReadOnlyList<string> lines)\` ([\`DaqifiStreamingDevice.cs:728\`](https://github.com/daqifi/daqifi-core/blob/main/src/Daqifi.Core/Device/DaqifiStreamingDevice.cs#L728)) checks the captured response from \`ExecuteTextCommandAsync\` for \`**ERROR\` / \`ERROR\` lines.

That tells the caller **a** command failed but not what's already in the device's error queue from prior commands or sessions. There is no \`DrainErrors\` / \`GetErrors\` method anywhere in the codebase — \`SYST:ERR?\` only appears in \`CompositeMessageParserTests.cs:173\` as a hardcoded test fixture, never in production code.

## Why this matters

This came up while validating a Nyquist (FW 3.4.6b1) where \`SYSTem:StartStreamData\` was returning \`**ERROR: -200, \"Execution error\"\` because the device was in Standby (\`SYSTem:POWer:STATe 0\`). Once we knew to wake it with \`POWer:STATe 1\`, streaming worked. But:

1. There were >10 stale errors queued from prior sessions; without explicit drain, every fresh \`SYST:ERR?\` query returned a stale error, not the current one
2. Even with the inline \`ContainsScpiError\` working correctly, the caller has no way to ask the device \"what errors do you have queued?\" for diagnostics or telemetry
3. A future tooling layer (CLI status report, automated regression test, Sentry breadcrumb) should be able to drain the queue and surface the errors as part of a health report

## Proposal

Add a method to \`DaqifiDevice\` (or a sibling diagnostics helper) that drains the device error queue and returns the popped entries:

\`\`\`csharp
/// <summary>
/// Pop SYST:ERR? entries from the device until the queue reports \"No error\".
/// </summary>
/// <param name=\"maxIterations\">Safety cap to avoid an infinite loop if the
/// reply parser misreads. Defaults to 256 — large enough to drain a deeply
/// queued device, small enough that a runaway loop is bounded.</param>
/// <returns>The list of error strings popped from the queue (empty if the
/// queue was already clean).</returns>
public async Task<IReadOnlyList<string>> DrainErrorQueueAsync(
    int maxIterations = 256,
    CancellationToken cancellationToken = default)
{
    var popped = new List<string>();
    for (int i = 0; i < maxIterations; i++)
    {
        cancellationToken.ThrowIfCancellationRequested();
        var lines = await ExecuteTextCommandAsync(
            () => Send(new ScpiMessage(\"SYST:ERR?\")),
            responseTimeoutMs: 200,
            cancellationToken: cancellationToken);
        // First non-empty line is the SCPI error response.
        var reply = lines.FirstOrDefault(l => !string.IsNullOrWhiteSpace(l)) ?? string.Empty;
        if (reply.Contains(\"No error\", StringComparison.OrdinalIgnoreCase))
        {
            return popped;
        }
        popped.Add(reply);
    }
    return popped; // caller can detect 'didn't converge' via Count == maxIterations
}
\`\`\`

Keep ownership of the popped entries with the caller so they can:
- Log them via \`Trace.WriteLine\` / a structured logger
- Surface them in a health-check report
- Throw if any indicate a hardware fault
- Discard if they're known-stale and just blocking forward progress

## Acceptance

- [ ] \`DrainErrorQueueAsync\` (or named equivalent) exists on \`DaqifiDevice\` with a configurable \`maxIterations\`
- [ ] Returns \`IReadOnlyList<string>\` of popped errors (does NOT discard)
- [ ] Uses \`ExecuteTextCommandAsync\` so it pauses the protobuf consumer correctly during streaming (otherwise the \`SYST:ERR?\` reply gets eaten)
- [ ] Unit tests cover: clean queue (returns empty), N queued errors (returns N), maxIterations cap (returns full list when not converged)
- [ ] XML doc comment notes the relationship to \`ContainsScpiError\` (drain is for queue inspection; ContainsScpiError is for last-command result)

## Refs

- Upstream pattern: inline \`ContainsScpiError\` in \`DaqifiStreamingDevice.cs:728\`
- Companion ticket on Python port: daqifi/daqifi-python-core (separate issue) — the Python port has a discard-only \`clear_errors()\` hardcoded to 10; it should mirror this returns-popped-errors design too

tylerkron:
Closes #174.

## Summary

- Adds `DaqifiDevice.DrainErrorQueueAsync` — pops `SYSTem:ERRor?` entries until the device reports `"No error"` and returns the popped strings to the caller. This complements the existing inline `ContainsScpiError` check (last-command result) by surfacing what's queued on the device, including stale errors from prior commands or sessions.
- Adds `ScpiMessageProducer.GetSystemError` (`SYSTem:ERRor?`) for use by the new method and any future callers.
- Adds 10 unit tests covering: clean queue, N queued errors, max-iterations cap, `+0,"No error"` terminator variant, leading blank lines, trimming, empty-reply termination, command verification, cancellation, and argument validation.

## Implementation notes (deviations from the issue's draft snippet)

- **Empty reply terminates the loop** instead of being recorded as an error. Without this, a transport timeout would cause the loop to append empty strings until `maxIterations` was exhausted.
- **Default response timeout matches `ExecuteTextCommandAsync`'s default (1000 ms)** rather than 200 ms — 200 ms was tight enough to cause false terminations on a slightly loaded transport.
- **When `maxIterations` is reached without convergence**, a `Trace.WriteLine` warning is emitted. Callers that want to treat non-convergence as a hard failure can compare `Count` to `maxIterations`.
- Method lives on `DaqifiDevice` (the base class) so all subclasses inherit it without duplication.

## Test plan

- [x] `dotnet test --filter DrainErrorQueue` — 10/10 passing
- [x] Full suite — 890/890 passing (2 pre-existing skips), no regressions

### DIFF
--- a/src/Daqifi.Core.Tests/Device/DaqifiDeviceDrainErrorQueueTests.cs
+++ b/src/Daqifi.Core.Tests/Device/DaqifiDeviceDrainErrorQueueTests.cs
@@ -75,6 +75,23 @@ namespace Daqifi.Core.Tests.Device
         }
 
         [Fact]
+        public async Task DrainErrorQueueAsync_TreatsErrorMessageContainingNoErrorPhraseAsRealError()
+        {
+            // Terminator detection must use the numeric SCPI code, not a
+            // substring match — otherwise a real error whose message text
+            // happens to contain "No error" would falsely terminate the drain.
+            var device = new SequencedReplyDevice("TestDevice");
+            device.Replies.Enqueue(new[] { "-200,\"Execution error: No error handler installed\"" });
+            device.Replies.Enqueue(new[] { "0,\"No error\"" });
+            device.Connect();
+
+            var popped = await device.DrainErrorQueueAsync();
+
+            Assert.Equal(new[] { "-200,\"Execution error: No error handler installed\"" }, popped);
+            Assert.Equal(2, device.ExecuteTextCommandCallCount);
+        }
+
+        [Fact]
         public async Task DrainErrorQueueAsync_OnEmptyReply_TerminatesEarly()
         {
             // Empty reply should NOT be treated as an error and should NOT cause
--- a/src/Daqifi.Core.Tests/Device/DaqifiDeviceDrainErrorQueueTests.cs
+++ b/src/Daqifi.Core.Tests/Device/DaqifiDeviceDrainErrorQueueTests.cs
@@ -1,0 +1,218 @@
+using Daqifi.Core.Communication.Messages;
+using Daqifi.Core.Device;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Daqifi.Core.Tests.Device
+{
+    public class DaqifiDeviceDrainErrorQueueTests
+    {
+        [Fact]
+        public async Task DrainErrorQueueAsync_WhenQueueIsClean_ReturnsEmpty()
+        {
+            var device = new SequencedReplyDevice("TestDevice");
+            device.Replies.Enqueue(new[] { "0,\"No error\"" });
+            device.Connect();
+
+            var popped = await device.DrainErrorQueueAsync();
+
+            Assert.Empty(popped);
+            Assert.Equal(1, device.ExecuteTextCommandCallCount);
+        }
+
+        [Fact]
+        public async Task DrainErrorQueueAsync_WhenSeveralErrorsQueued_ReturnsAllInOrder()
+        {
+            var device = new SequencedReplyDevice("TestDevice");
+            device.Replies.Enqueue(new[] { "-200,\"Execution error\"" });
+            device.Replies.Enqueue(new[] { "-113,\"Undefined header\"" });
+            device.Replies.Enqueue(new[] { "-410,\"Query INTERRUPTED\"" });
+            device.Replies.Enqueue(new[] { "0,\"No error\"" });
+            device.Connect();
+
+            var popped = await device.DrainErrorQueueAsync();
+
+            Assert.Equal(new[]
+            {
+                "-200,\"Execution error\"",
+                "-113,\"Undefined header\"",
+                "-410,\"Query INTERRUPTED\"",
+            }, popped);
+            Assert.Equal(4, device.ExecuteTextCommandCallCount);
+        }
+
+        [Fact]
+        public async Task DrainErrorQueueAsync_AcceptsPlusZeroNoErrorTerminator()
+        {
+            var device = new SequencedReplyDevice("TestDevice");
+            device.Replies.Enqueue(new[] { "-200,\"Execution error\"" });
+            device.Replies.Enqueue(new[] { "+0,\"No error\"" });
+            device.Connect();
+
+            var popped = await device.DrainErrorQueueAsync();
+
+            Assert.Single(popped);
+            Assert.Equal("-200,\"Execution error\"", popped[0]);
+        }
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


tylerkron:
Replying to @qodo-code-review's review:

**Issue 1 — Loose terminator detection: fixed in [b4fc3ad](https://github.com/daqifi/daqifi-core/pull/185/commits/b4fc3ad).**

Agreed. Replaced the `Contains("No error")` substring match with numeric-code parsing: parse the prefix up to the first comma and only treat code `0` (or `+0`, via `int.TryParse`) as the terminator. Added a regression test (`DrainErrorQueueAsync_TreatsErrorMessageContainingNoErrorPhraseAsRealError`) that fails under the old behavior — it queues a real error whose message text contains "No error handler installed" and asserts it's captured rather than swallowed as the terminator.

**Issue 2 — Unsynchronized text exchange: valid concern, but out of scope for this PR. Tracked as [#186](https://github.com/daqifi/daqifi-core/issues/186).**

The lack of locking is in `ExecuteTextCommandAsync` itself, which predates this PR. Every existing method that uses it (`InitializeAsync`, `GetSdCardFilesAsync`, `DeleteSdCardFilesAsync`, `DownloadSdCardFileAsync`, plus the SD-card paths in `DaqifiStreamingDevice`) shares the same vulnerability. The right fix is a `SemaphoreSlim` inside `ExecuteTextCommandAsync` so all callers benefit — bolting one onto `DrainErrorQueueAsync` alone would still leave the other paths exposed. Filed [#186](https://github.com/daqifi/daqifi-core/issues/186) to do this properly across the board.

tylerkron:
## Real-hardware verification

Tested against a Nyquist Nq1 FW 3.4.4 over USB serial (`/dev/cu.usbmodem101` @ 115200).

### What worked
- Connect + `InitializeAsync` + drain run cleanly end-to-end
- Multiple sequential `DrainErrorQueueAsync` calls — protobuf consumer pauses/restarts correctly each time, no transport corruption
- Raw `SYSTem:ERRor?` reply on a clean queue is exactly `0,"No error"` — the numeric-prefix terminator parsing introduced for Qodo issue #1 hits the right pattern

### Firmware caveat (FW 3.4.4)
On this firmware, **errors are reported inline only and are NOT enqueued in `SYST:ERR?`**. Every bad command I tried produced an inline `**ERROR:` line but left the queue at `0,"No error"`:

| Command | Inline reply | Queue after |
|---|---|---|
| `SYSTem:DOES:NOT:EXIST?` | `**ERROR: -113, "Undefined header"` | `0,"No error"` |
| `SYSTem:POWer:STATe 99` | `**ERROR: -224, "Illegal parameter value"` | `0,"No error"` |
| `SD:GET "nonexistent.bin"` | `**ERROR: -200, "Execution error"` | `0,"No error"` |
| Standby + StartStreamData (exact issue-174 reproducer) | `**ERROR: -200, "Execution error"` | `0,"No error"` |

The original issue mentioned `>10 stale errors queued` on FW 3.4.6b1, so queue-population is a newer firmware behavior. `DrainErrorQueueAsync` will return empty on 3.4.4 but should populate correctly on firmware that implements the queue (terminator parsing already verified against the standard `0,"No error"` reply shape).

### What I couldn't prove on this hardware
- "N popped entries when N are queued" — would need 3.4.6b1+ to seed and verify

The mechanical primitives all work; the populated-list path is exercised by the unit tests.

PR #347, opened 2026-07-18
feat(sd): allow SD file transfer over WiFi on firmware >= v3.7.0

cptkoolbeenz:
## Summary

SD file operations (LIST / GET / DELETE) were hard-gated to USB by a blanket `if (!IsUsbConnection) throw` on `DownloadSdCardFileAsync`, citing SD/WiFi shared-SPI-bus contention. That contention was real — **but only until firmware v3.7.0**: firmware **#598/#599** (commit `bf105585`, first released tag **v3.7.0** per `git tag --contains`) route the SD reply to the requesting interface, making SD-over-WiFi safe on v3.7.0+.

This replaces the transport-only block with a **firmware-capability gate**, following the existing [ADR 0001](docs/adr/0001-firmware-feature-gating.md) framework.

## Changes

- **`DeviceFeature.SdFileTransferOverWifi`** — min firmware v3.7.0; needs WiFi + SD HW.
- **`EnsureSdFileTransferSupportedOnTransport()`** — over USB, always allowed on SD-capable firmware; over WiFi/TCP, requires firmware ≥ v3.7.0 (unparseable/older reported version → unsupported) and throws a typed `FeatureNotSupportedException(SdFileTransferOverWifi, v3.7.0, actual, board)` **up front** rather than dispatching a command the firmware can't service over WiFi (which would stall on the shared SPI bus).
- Applied to **`GetSdCardFilesAsync`** (LIST), **`DownloadSdCardFileAsync`** (GET), **`DeleteSdCardFileAsync`** (DELETE) — all three depend on #598's reply-routing. LIST was previously ungated over WiFi (silent stall on old firmware); now gated too.
- **ADR 0001** living-evidence updated: timeline extended through v3.7.2; SD-over-WiFi row added; the "no live above-floor gate today" note corrected — **this is the first active above-floor firmware-version gate**.

## Behavior change

Over WiFi on old/unknown firmware, these operations now throw the typed `FeatureNotSupportedException` instead of (download) the blanket USB-only `InvalidOperationException` or (list) a silent stall. **Over USB: unchanged** (available on all SD-capable firmware).

## Tests

New SD-over-WiFi gate coverage: at / above / below min + unparseable version; over USB not gated; LIST / GET / DELETE. Existing non-USB download test updated to the typed exception; `RetryableSdCardStreamingDevice` marked USB (its LIST retry/parse tests are transport-agnostic). **Full suite: 1398 passed / 0 failed / 2 skipped** (net10.0).

## Context

Enables [daqifi-avalonia #1](https://github.com/daqifi/daqifi-avalonia/issues/1) (SD offload over WiFi) — the app-side USB gate removal ([daqifi-avalonia PR #2](https://github.com/daqifi/daqifi-avalonia/pull/2)) was necessary but not sufficient; this Core guard was the other USB-only policy. Consumers pinning the NuGet package need a release to pick this up.

**Note:** the v3.7.0 threshold is derived from firmware git history (`git describe bf105585` = `v3.6.3-6`, first containing tag v3.7.0). On-device `SD:LIST?` + `SD:GET` over TCP validation against a v3.7.x device is still recommended before release.


tylerkron (approved):
Approving as code-owner (not the author). Basis: green CI, 7 rounds of Qodo, full unit coverage; verified locally against latest `main` (merge clean; 523 SD/WiFi/feature-gate tests pass on net9+net10) and bench-confirmed on live Nq1 (FW 3.7.2) that the **USB** SD path is unchanged (`--sd-list` returns files; the new transport gate does not block USB).

⚠️ **Caveat:** the **live SD-over-WiFi/TCP path was NOT validated** — the bench board's WINC module is up over USB (ChipId 1377184, FW 19.7.7) but it isn't reachable on the LAN (no UDP discovery response; nothing answering TCP 9760 on the subnet). Merging on user authorization + code/CI/Qodo strength; on-device `SD:LIST?`/`SD:GET` over TCP against a networked v3.7.x board remains a recommended follow-up.

PR #547, opened 2026-08-15
fix(producers): a producer mid-write could report itself idle

tylerkron:
## What was wrong

`MessageProducer.IsIdle` answers "nothing queued **and** no write part-way out to the stream". It could answer that wrongly: it would report **idle while a write was still in flight**.

That is the exact false "all quiet" the property exists to rule out, and production code acts on it. `OperationSerializer.DrainOutboundQueueAsync` waits on `IsIdle` before swapping the stream's reader, and its own docs explain why it can't just check the queue count:

> A message is taken off the queue *before* it is written, so the count reads zero while the producer thread is still inside a blocking write.

A phantom idle defeats that barrier — it can swap the reader mid-write and collect the outgoing command's reply as part of its own exchange (the failure mode of #342).

## Why it happened

`IsIdle` is **two separate field reads**, and they were in the order that lets a caller straddle the background loop's handover from *queued* to *being written*:

1. reader reads `_draining` → `false` (the loop has woken but not yet claimed the batch)
2. the loop claims the batch (`_draining = true`) and dequeues the message
3. reader reads `IsEmpty` → `true`
4. → `IsIdle` returns `true`, with the write in flight

Claiming `_draining` before the dequeue was only half the guarantee. The other half is the order `IsIdle` reads the two fields in.

## The fix

```diff
-public bool IsIdle => !_draining && _messageQueue.IsEmpty;
+public bool IsIdle => _messageQueue.IsEmpty && !_draining;
```

Whichever side of the handover the two reads land on, one of them now still reports work outstanding. No behaviour change other than removing the false positive; the conservative direction (briefly reporting busy while idle) was already possible and is safe.

## How it surfaced

As an intermittent full-suite failure in `MessageProducerIdleWakeupTests` — `WakeCount_IsCumulativeAcrossRestarts` and `AfterDrainingAMessage_TheProducerGoesBackToNotWaking`, both failing in the shared `WaitUntilIdle` helper with "The producer never drained its queue."

The helper polled `IsIdle` and then asserted on a *second* read, so a phantom idle ended the poll and the re-read failed it. The tell was that it failed in **under a millisecond** — a starved or stalled producer would have burned the full deadline. That ruled out the timing explanation and pointed at the read.

## Evidence

Sampling the handover tightly (spin instead of a 5 ms poll — the poll interval is the only reason this surfaced rarely):

| | phantom idles |
|---|---|
| before the fix | **498 / 20,000** (~2.5%) |
| after the fix | 0 / 20,000 |
| after the fix | 0 / 500,000 |

Full-suite verification: **20 consecutive `dotnet test -f net10.0` runs with no `MessageProducerIdleWakeupTests` failure.** Before the fix it recurred roughly 1 run in 10.

## Test changes

- **New regression test** `Producer_AtTheHandoverFromQueuedToWriting_NeverReportsAPhantomIdle`, placed beside the existing `Producer_WithAWriteInFlight_ReportsQueueEmptyButNotIdle`, which pins the same guarantee in steady state and cannot see this race. Confirmed it fails on the old code (498 phantoms) and passes on the new (0). Runs in ~0.7 s.
- **`WaitUntilIdle` now decides on a single `IsIdle` read** rather than polling on one and asserting on another. The wait is otherwise unchanged — the producer must still reach idle within 5 s or the test fails.

Nothing that `WakeCount_IsCumulativeAcrossRestarts` asserts about cumulative wake counts was weakened; it still requires the count to carry across a restart and grow. That assertion never actually failed — across 60,000 stress iterations it was always the shared helper.

## Note on #531

Not the same family. #531 is an allocation-measurement flake (`GC.GetAllocatedBytesForCurrentThread` on a shared xUnit worker); this was a genuine data race in `src/`. #531 stays open.

A third, unrelated intermittent turned up during verification and is **not** addressed here: `LiveSampleStreamTests.ConcurrentEnumerations_AllEndOnTheSameDrop` fails ~2 runs in 20 (subscriber count 1, expected 0) — it looks like the assertion racing the enumerators' unsubscribe.

### DIFF
--- a/src/Daqifi.Core.Tests/Communication/Producers/MessageProducerIdleWakeupTests.cs
+++ b/src/Daqifi.Core.Tests/Communication/Producers/MessageProducerIdleWakeupTests.cs
@@ -19,6 +19,17 @@ public class MessageProducerIdleWakeupTests
     /// </summary>
     private const int IdleObservationMs = 400;
 
+    /// <summary>
+    /// Gap between checks of <see cref="MessageProducer{T}.IsIdle"/> in <see cref="WaitUntilIdle"/>.
+    /// </summary>
+    private const int IdlePollIntervalMs = 5;
+
+    /// <summary>
+    /// How long <see cref="WaitUntilIdle"/> waits before calling the producer hung. The drain lands
+    /// on the first or second poll in practice; this is only ever spent by a test that is failing.
+    /// </summary>
+    private const int IdleDrainTimeoutMs = 5000;
+
     [Fact]
     public void AnIdleProducer_NeverWakes()
     {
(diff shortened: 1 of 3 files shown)


cptkoolbeenz:
## Both findings applied — and one change removes the cause of both

**Stress test spawns 20k threads.** Right, and the loop did not need to. It created a fresh `MessageProducer` per attempt and `Start()` creates a `Thread` each time, so 20,000 attempts meant 20,000 threads for no extra coverage: the background loop returns to waiting on `_messageAvailable` after each drain, so **one** producer sent to many times re-enters exactly the queued → writing handover this test exists to sample. It is one producer now.

**Unbounded idle spin loop.** Also right, and the worse of the two — a regression that stopped the producer idling would have hung CI indefinitely burning a core rather than failing with a name. The spin now asserts against `DeadlockBudget` and says which bug it is *not*:

> The producer never reported idle within 15s on attempt N. Something is keeping it draining, which is a different bug from the phantom idle this test samples for.

**And the count became a clock.** A fixed attempt count is the wrong knob for a race sampler — it makes a slow CI runner pay for the fast one's sampling density. It runs for 2 s and takes as many samples as fit, with a floor (`attempts > 100`) so a loop that quietly became a no-op cannot pass silently.

The spin itself stays a spin: the window is a handful of instructions wide and a poll that sleeps between reads steps straight over it. That was never the problem; the missing deadline was.

**Result:** 29 tests in this file pass in **7 s** on net10.0, against a suite that previously spent most of that one test's runtime creating threads.

PR #550, opened 2026-08-19
feat(sd): understand the device's end-of-listing marker

cptkoolbeenz:
Client half of firmware [PR #796 — tell the host where a directory listing ended](https://github.com/daqifi/daqifi-nyquist-firmware/pull/796) (firmware issue [#794](https://github.com/daqifi/daqifi-nyquist-firmware/issues/794)). **The firmware PR does not merge until this one does** — see "Why this must land first".

## What the firmware adds

`SYSTem:STORage:SD:LISt?` now ends with one marker line:

| marker | meaning |
|---|---|
| `__END_OF_LIST__ OK` | walked the whole tree, every entry emitted |
| `__END_OF_LIST__ INCOMPLETE` | finished, but entries were skipped (depth cap, unreadable subdirectory, a path that did not fit) |
| `__END_OF_LIST__ FAILED` | nothing could be listed at all |

An **aborted** listing sends no marker at all, deliberately: the abort happens because the peer stopped draining, and each further chunk costs ~10 s of transfer retries.

## Why this must land first

`SdCardFileListParser.ParseFileList` takes the **first token** of every non-blank, non-error line and turns it into a file:

```csharp
var tokenEnd = path.IndexOfAny(new[] { ' ', '\t' });
if (tokenEnd > 0) { path = path.Substring(0, tokenEnd); }
var fileName = Path.GetFileName(path);
if (string.IsNullOrWhiteSpace(fileName)) { continue; }
files.Add(new SdCardFileInfo(fileName, createdDate));
```

`__END_OF_LIST__ OK` is not blank and is not error-shaped, so without this change the desktop app would list a **phantom file called `__END_OF_LIST__`** that the user can neither open nor delete.

## What this does

1. **Skips the marker** in `ParseFileList` — required for the above.
2. **Uses it.** New public `SdCardListingStatus` + `SdCardFileListParser.GetListingStatus(lines)`. It reads the **last** marker in the reply: one leading the response is a stale marker from a timed-out exchange, the same reasoning `TrySplitAtSdListTerminator` already applies a few lines away.
3. `GetSdCardFilesAsync` treats **FAILED** as an incomplete listing (`SdCardListIncompleteException`) instead of returning an empty list. An empty list from a walk that could not open the directory reports *"there is nothing there"* for what was really *"I could not look"* — the confusion this mechanism exists to end.

`Unterminated` is **not** an error: firmware predating #796 sends no marker, and neither does an aborted listing. Callers simply learn nothing more from the device, which is where we already are today. An **unknown** status word counts as `Incomplete`, not `Complete` — a future firmware saying something this version does not understand has still told us not to trust the list as whole.

## Note on what was already here

`TrySplitAtSdListTerminator` already gives this client a completeness signal by chasing the listing with `SYST:ERR?` and requiring its reply to arrive after the entries. That is a genuine and clever guard, and it is why "the host cannot tell" was too strong a claim in the firmware issue. What it cannot do is distinguish a walk that **skipped** part of the card from one that saw all of it, or an unopenable directory from an empty one — the device is the only thing that knows that, and now it says so.

## Test plan

- 6 new unit tests: marker not parsed as a file; each status word; no marker → `Unterminated`; unknown word → `Incomplete`; a stale leading marker loses to the trailing one.
- `dotnet test`: **43 passed, 0 failed** on net9.0 and net10.0.
- **Mutation-proved**: disabling the skip line fails exactly one test (43 → 42 passed on both frameworks), so the marker test cannot pass vacuously.
- Hardware: covered by the firmware PR's bench run and by `test_794_sd_list_terminator.py` in daqifi-python-test-suite ([#151](https://github.com/daqifi/daqifi-python-test-suite/pull/151)).

### DIFF
--- a/src/Daqifi.Core.Tests/Device/SdCard/SdCardFileListParserTests.cs
+++ b/src/Daqifi.Core.Tests/Device/SdCard/SdCardFileListParserTests.cs
@@ -24,6 +24,34 @@ namespace Daqifi.Core.Tests.Device.SdCard
             Assert.Equal("data.csv", result[0].FileName);
         }
 
+        [Fact]
+        public void ParseFileList_WithFileNamedLikeTheMarker_KeepsTheFile()
+        {
+            // The marker is matched as a whole token, not as a prefix: a file
+            // whose name merely starts with the marker text is still a file, and
+            // hiding it would be the same defect as inventing one.
+            // No "Daqifi/" prefix on purpose: with one, the line does not start
+            // with the marker text at all and the exactness guard is never
+            // reached -- the test would pass whether or not the guard exists.
+            var lines = new[] { "__END_OF_LIST__notes.csv 12", "__END_OF_LIST__ OK" };
+
+            var result = SdCardFileListParser.ParseFileList(lines);
+
+            Assert.Single(result);
+            Assert.Equal("__END_OF_LIST__notes.csv", result[0].FileName);
+        }
+
+        [Fact]
+        public void GetListingStatus_WithFileNamedLikeTheMarker_IsUnterminated()
+        {
+            // Same rule from the other side: a filename starting with the marker
+            // text must not be read as the listing's terminator.
+            var status = SdCardFileListParser.GetListingStatus(
+                new[] { "__END_OF_LIST__notes.csv 12" });
+
+            Assert.Equal(SdCardListingStatus.Unterminated, status);
+        }
+
         [Theory]
         [InlineData("__END_OF_LIST__ OK", SdCardListingStatus.Complete)]
         [InlineData("__END_OF_LIST__ INCOMPLETE", SdCardListingStatus.Incomplete)]
--- a/src/Daqifi.Core.Tests/Device/SdCard/SdCardFileListParserTests.cs
+++ b/src/Daqifi.Core.Tests/Device/SdCard/SdCardFileListParserTests.cs
@@ -66,6 +66,22 @@ namespace Daqifi.Core.Tests.Device.SdCard
         }
 
         [Fact]
+        public void GetListingStatus_WithContentAfterTheMarker_IsUnterminated()
+        {
+            // A marker with entries after it did not terminate THIS listing --
+            // it is left over from an earlier, timed-out exchange. Claiming
+            // Complete there would vouch for a reply we never saw the end of.
+            var status = SdCardFileListParser.GetListingStatus(new[]
+            {
+                "Daqifi/a.csv 1",
+                "__END_OF_LIST__ OK",
+                "Daqifi/b.csv 2",
+            });
+
+            Assert.Equal(SdCardListingStatus.Unterminated, status);
+        }
+
+        [Fact]
         public void GetListingStatus_WithNoMarker_IsUnterminated()
         {
             // Pre-#794 firmware, and the abort case, which deliberately sends none.
--- a/src/Daqifi.Core.Tests/Device/SdCard/SdCardFileListParserTests.cs
+++ b/src/Daqifi.Core.Tests/Device/SdCard/SdCardFileListParserTests.cs
@@ -7,6 +7,72 @@ namespace Daqifi.Core.Tests.Device.SdCard
 {
     public class SdCardFileListParserTests
     {
+        // ---- firmware #794: the end-of-listing marker -----------------------
+
+        [Fact]
+        public void ParseFileList_WithEndOfListMarker_DoesNotYieldAFile()
+        {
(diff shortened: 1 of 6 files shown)


cptkoolbeenz:
Both open suggestions are already applied — noting the commits since the checkboxes have not flipped:

- **"Reject explicitly partial listings"** (9) → `c3ea540`. `INCOMPLETE` joins `FAILED` in raising `SdCardListIncompleteException`: the caller is about to cache the list and answer "is my file on the card?" from it, and a device-truncated listing is the same class of wrong answer as the transport-truncated one the method already rejects.
- **"Validate that markers terminate listings"** (7) → `efc8fe1`. Content after a marker resets the status to `Unterminated` — a marker with entries following it did not terminate *this* listing, it was left in flight by an earlier timed-out exchange. Mutation-proved (46 → 45 passed with the reset disabled).

Since then, from the adversarial audit: `ed75300` gives both delete-path `GetSdFileList` exchanges the same 1000 ms completion window the read path uses. Without it they took the 250 ms default, so a merely-slow listing was cut off *before* its terminator, read as `Unterminated` — which by design does not throw — and slipped past the new guard to overwrite the cache with a partial list. The guard can only judge a marker it was given time to receive.

One note on that fix worth recording: my first pass patched the `GetSdSpace` exchange too, by pattern. Reverted before pushing — a two-number reply has nothing to walk, and the comment I had attached to it was simply false there.

506 SD tests pass on net9.0 and net10.0.

cptkoolbeenz:
## `4554e34` — the delete relist now retries a stall, like the read path

Requiring the transport terminator (last commit) made a *merely slow* listing fatal: the retry block was gated only on `ContainsScpiError`, so an unterminated-but-error-free reply skipped it entirely and threw on a delete the device had already carried out. The read path has always treated that as a one-off stall and given it a second attempt — its own comment says so — and the two paths should not disagree about the same transient.

**The retry re-LISTs only.** Re-sending the `DELETE` would ask the device to remove a file that is already gone, turn a transient stall into a hard error, and exhaust the retry anyway. That is why it is a separate loop from the error retry above, which re-sends both deliberately: there the delete itself may not have landed.

Two tests, and one of them had to be hardened: the existing throw case used `UnterminatedAttempts = 1`, which the retry now recovers from — it would have proved the retry works rather than the throw. It withholds the terminator on every attempt now. That is the second time on this PR that a test passed for a reason other than the one it claimed, both caught by mutation-proving; the first was the fake quietly appending the terminator I had deliberately left out.

**508 SD tests pass** on net9.0 and net10.0 (the commit message says 509 — miscount, the run output is 508). Mutation-proved: 25 → 24 with the retry disabled.

### Where this PR stands

Five audit rounds, each finding one real thing, all in the same small area and each narrower than the last:

1. the marker became a phantom file in `ParseFileList`
2. the status was read but not *used* at the second call site
3. the delete exchange took the 250 ms window, so the marker often never arrived
4. a marker-less reply still overwrote the cache — the transport terminator was missing
5. requiring that terminator made a stall fatal where the read path retries

Nothing outstanding.

cptkoolbeenz:
## Triage of the current suggestion set

**Applied** — *Verify complete listing before suppression* / *Validate that markers terminate listings* / *Reject explicitly partial listings*, and the `/agentic_review` bug **Unterminated listing masks delete error**. These are one finding and it was right. `goneFromCard` read the transport terminator as proof the device had finished walking the card; it only says the reply arrived intact. A walk the device cut short — `INCOMPLETE`, or an `Unterminated` listing from firmware that sends no marker — is missing entries it never reached, so the target's absence proved nothing, and a delete the device refused came back as success. It now requires the device's own `Complete` marker (`3024fe5`). Against firmware without a marker the error stands, which is the behaviour that predates this check: outcome verification is a capability the #794 terminator enables, not a weakening of the error path. Mutation-proved 517 → 515.

**Declined** — *Preserve files matching the marker* (importance 7). The premise needs a listing line whose **first token** is exactly `__END_OF_LIST__`. The firmware cannot emit one: every entry is built as `snprintf(newPath, …, "%s/%s", <dir>, stat.fname)` (`sd_card_manager.c`), so a file actually named `__END_OF_LIST__` arrives as `Daqifi/__END_OF_LIST__ 1234` — first token `Daqifi/__END_OF_LIST__`, which does not start with the marker and is already kept as a file. Even an empty directory setting yields `/__END_OF_LIST__`. The paired idea of rejecting numeric status words has a cost on the other side: it would turn any future numeric status into a phantom file rather than an unrecognised marker, and the wire format has no case that needs it.

Nothing outstanding in the blocking tier (`review_bugs=0`).

Issue #538, opened 2026-08-14
perf(diagnostics): three log calls stall 2.5-3s holding the operation lock — a bare-LF reply isn't recognised, and an empty log waits out the full timeout


tylerkron:
## What goes wrong

Three of the diagnostics calls take **2.5–3.1 seconds** instead of the ~1.05 s the rest of them take, and they hold the device operation lock for the whole time — so streaming control, SD work and every other text command queue up behind them.

The three are `ClearSystemLogAsync`, `TestSystemLogAsync`, and — the one that actually matters — `GetSystemLogAsync` **whenever the log buffer is empty**, which on a healthy device is the normal case. Polling a quiet device's log costs 3 seconds a go.

## Why — two different causes that look identical

**1. A bare `\n` doesn't end a line.** Most firmware replies are CRLF-terminated, but two of them are not, and those two are exactly the slow ones:

```
SYSTem:ERRor:COUNt?   ->  b'0\r\n'                        Core: 1056 ms
SYSTem:LOG?           ->  b'Test error message\r\n\r\n'   Core: 1060 ms
SYSTem:LOG:CLEar      ->  b'Log cleared\n'                Core: 3068 ms   <-- bare LF
SYSTem:LOG:TEST       ->  b'Added test log messages\n'    Core: 2561 ms   <-- bare LF
```

The correlation is exact. The device answers immediately in all four cases; Core just doesn't recognise a bare-LF reply as a completed line, so it waits out the full `responseTimeoutMs` (2000 ms, set by `DeviceDiagnosticsOperations.DIAGNOSTICS_RESPONSE_TIMEOUT_MS`) before giving up and surfacing the content anyway on teardown. Nothing fails — it is purely dead time.

**2. An empty log buffer produces no bytes at all.** `LogMessageDump()` is a pop-and-print loop over the ring buffer; with nothing buffered it writes *zero bytes* and returns. So `GetSystemLogAsync` on an empty log has nothing to wait for and always burns the full 2 s timeout.

That second one has a correctness edge to it as well: an empty log and a device that simply didn't answer are **indistinguishable** — both return an empty list after ~3 s. `IsErrorOnlyResponse` correctly declines to throw on a no-line response (verified), so the "empty log vs command failed" distinction the code comments reason about only covers the case where the device answers *with an error*, not the case where it answers with nothing.

## Evidence

[...]

PR #552, opened 2026-08-20
fix(diagnostics): tell an empty system log apart from a device that never answered

### LINKED ISSUE #543: bug(diagnostics): an empty system log and a device that never answered are indistinguishable — but the firmware sends a terminator that would separate them
## What goes wrong

`GetSystemLogAsync` returns an empty list in two completely different situations:

- the device answered, and its log buffer is genuinely empty (the normal case on a healthy device);
- the device did not answer at all, and the exchange gave up after its first-response timeout.

A caller cannot tell which happened. `ThrowIfErrorOnlyResponse` only covers the third case — the device answering *with an error* — so a silent link, a wedged text exchange or an unsupported header on below-floor firmware all present as "your log is empty".

This was called out as the correctness edge of the second half of #538, and deliberately left out of scope there: #538 was a framing/latency fix, and changing what this method throws is a public behaviour decision on its own.

## Why it is now cheap to fix

The bench work for #538 established what the firmware actually sends, which the original issue had wrong. An empty `SYSTem:LOG?` does **not** produce zero bytes — it answers `\r\n`, in 6 ms, measured three trials out of three on a bench Nq1 running firmware 3.7.2 with a raw pyserial probe (no Core in the path). A populated dump answers its entries followed by the same trailing `\r\n`:

```
SYSTem:LOG?   (empty)      ->  b'\r\n'
SYSTem:LOG?   (populated)  ->  b'Test error message\r\n\r\n'
SYSTem:LOG?   (3261 bytes) ->  b'...\r\nTest error message\r\n\r\n'
```

So **the trailing blank line is an end-of-dump terminator the firmware already sends unconditionally.** A log read that sees it knows the device answered; one that does not see it knows the device went silent. No extra command, no `SYSTem:ERRor?` pairing, and — unlike that option, which the SD listing uses — no popping an entry off the device's error queue as a side effect, which would be a poor thing for a diagnostics read to do given `GetSystemErrorCountAsync` sits on the same interface.

The exchange already surfaces the blank line internally as of #538 (`LineBasedMessageParser.EmitEmptyLines`); it filters it out of the result before callers see it, so the signal exists and is simply not routed anywhere yet.

## Worth deciding

The mechanism is settled; the API shape is not.

- **What should a missing terminator do?** Throwing a `DeviceDiagnosticsException` is the honest answer and matches how the SD listing treats its missing terminator, but it turns a call that used to return empty into one that throws — worth a moment's thought for anyone polling the log on a flaky link.
- **Should it be narrowed to `SYSTem:LOG?` only,** or does `GetCommandHistoryAsync` have an equivalent? That one has a "No command history" marker, so it may already be distinguishable and need nothing.
- The false-rejection risk is a lost terminator on an otherwise good response, which would report a failure that did not happen. Worth a bench check that the blank line survives a large dump (the 3261-byte one above did, with a max inter-chunk gap of 8 ms).

## Scope

Small and self-contained: one guard in `DeviceDiagnosticsOperations.GetSystemLogAsync`, plumbing the terminator out of `TextExchangeEngine` (or letting the log read opt into seeing blank lines), and tests. Bench-validatable non-destructively — clearing and re-populating the volatile log buffer is the whole experiment.

cptkoolbeenz:
Fixes #543.

## The defect

`GetSystemLogAsync` returned an empty list for two unrelated situations:

- the device answered, and its log buffer is genuinely empty;
- the device **did not answer at all**.

A caller could not tell which. So a silent link, a wedged text exchange, or an unsupported header on below-floor firmware all presented as *"your log is empty"* — the least useful answer a **diagnostics** call can give, because it is indistinguishable from a clean bill of health, on the one call an operator reaches for when they suspect the device is unwell.

## The signal was already there

The firmware terminates **every** `SYSTem:LOG?` dump with a blank line, whether or not it had anything to say. From #543's bench work (Nq1 on firmware 3.7.2, raw pyserial probe, no Core in the path, three trials of three):

```
SYSTem:LOG?  (empty)      ->  b'\r\n'                        6 ms
SYSTem:LOG?  (populated)  ->  b'Test error message\r\n\r\n'
```

Since #538 the exchange already parses that blank line internally — it just filtered it out before any caller saw it. So **any** line reaching the diagnostics layer means the device answered, and **zero** lines means it did not.

No extra command, and specifically **no `SYSTem:ERRor?` pairing**, which would pop an entry off the device's error queue as a side effect of a read — a poor thing for a diagnostics call to do when `GetSystemErrorCountAsync` sits on the same interface.

## Why a parameter and not a new method

`keepBlankLines` is threaded through the existing `ExecuteTextCommandAsync` rather than added as a parallel method — the convention this seam already documents:

> *"a parallel method would be bypassed silently by any subclass that overrides only this one … Overriders must widen their signature — a compile error, which is the point."*

**That is most of this diff: 22 test doubles gain one parameter.** It is also load-bearing rather than incidental — the diagnostics tests inject their canned responses by overriding that virtual, so a new seam would bypass their injection and the behaviour could not be unit-tested at all. I checked `ExecuteRawCaptureAsync` as an alternative first; using it would mean reimplementing line reading, timeouts and consumer swapping in the diagnostics op, which is worse.

Only the `Action` overload is widened — the 8 overrides of the async overload are untouched.

## A test that encoded the wrong belief

`GetSystemLogAsync_WhenBufferEmpty_ReturnsEmpty` canned **nothing** and asserted no-throw, commented *"No lines = genuinely empty buffer (firmware writes nothing)"*.

That premise is false, and it is precisely what made the two states indistinguishable — the test asserted the ambiguity as correct. It now cans a **blank line**, which is what the device actually sends, and two cases join it:

| test | asserts |
|---|---|
| `WhenBufferEmpty_ReturnsEmpty` | a blank line → empty list, **no throw** |
| `WhenDeviceAnswersNothingAtAll_Throws` | zero lines → `DeviceDiagnosticsException` naming the condition |
| `WhenBufferHasEntries_IgnoresTheTerminator` | a populated dump does not leak its terminator as a phantom log entry |

## Behaviour change — worth a release note

A caller polling the log on a flaky link previously got an empty list and now gets a throw. That is the point of the issue, but it is visible. The exception is a `DeviceDiagnosticsException` (not a new type), so `catch` blocks that already handle diagnostics failures keep working.

## Test plan

- **Baseline-checked, not assumed.** `main` fails **4** tests on this station before any change — all `Device.Discovery.LinuxUsbPortDescriptorProviderTests`, a WSL/Windows runtime artefact, not related to this work.
- **After this change: the same 4**, with passing tests up from **3531 → 3533** (the two new cases) and total 3537 → 3539.
- `Daqifi.Mcp.Tests` builds clean (its `DeviceDouble` is one of the widened doubles).
- No hardware needed — the mechanism is pinned by unit tests against the real response shapes.

The issue's open questions, answered: it is scoped to `SYSTem:LOG?` only. `GetCommandHistoryAsync` is untouched — it has a *"No command history"* marker and is already distinguishable, exactly as the issue suspected.

### DIFF
--- a/src/Daqifi.Core.Tests/Device/Capabilities/DaqifiDeviceCapabilityDocumentTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Capabilities/DaqifiDeviceCapabilityDocumentTests.cs
@@ -306,7 +306,8 @@ public class DaqifiDeviceCapabilityDocumentTests
             int completionTimeoutMs = 250,
             CancellationToken cancellationToken = default,
             Func<CancellationToken, Task>? prepareAsync = null,
-            Func<Task>? finalizeAsync = null)
+            Func<Task>? finalizeAsync = null,
+            bool keepBlankLines = false)
         {
             try
             {
--- a/src/Daqifi.Core.Tests/Device/DaqifiDeviceDrainErrorQueueTests.cs
+++ b/src/Daqifi.Core.Tests/Device/DaqifiDeviceDrainErrorQueueTests.cs
@@ -209,7 +209,8 @@ namespace Daqifi.Core.Tests.Device
                 int completionTimeoutMs = 250,
                 CancellationToken cancellationToken = default,
                 Func<CancellationToken, Task>? prepareAsync = null,
-                Func<Task>? finalizeAsync = null)
+                Func<Task>? finalizeAsync = null,
+                bool keepBlankLines = false)
             {
                 try
                 {
--- a/src/Daqifi.Core.Tests/Device/DaqifiDeviceInitializeTests.cs
+++ b/src/Daqifi.Core.Tests/Device/DaqifiDeviceInitializeTests.cs
@@ -539,7 +539,8 @@ namespace Daqifi.Core.Tests.Device
                 int completionTimeoutMs = 250,
                 CancellationToken cancellationToken = default,
                 Func<CancellationToken, Task>? prepareAsync = null,
-                Func<Task>? finalizeAsync = null)
+                Func<Task>? finalizeAsync = null,
+                bool keepBlankLines = false)
             {
                 try
                 {
@@ -654,7 +655,8 @@ namespace Daqifi.Core.Tests.Device
                 int completionTimeoutMs = 250,
                 CancellationToken cancellationToken = default,
                 Func<CancellationToken, Task>? prepareAsync = null,
-                Func<Task>? finalizeAsync = null)
+                Func<Task>? finalizeAsync = null,
+                bool keepBlankLines = false)
             {
                 try
                 {
@@ -763,7 +765,8 @@ namespace Daqifi.Core.Tests.Device
                 int completionTimeoutMs = 250,
                 CancellationToken cancellationToken = default,
                 Func<CancellationToken, Task>? prepareAsync = null,
-                Func<Task>? finalizeAsync = null)
+                Func<Task>? finalizeAsync = null,
+                bool keepBlankLines = false)
             {
                 try
                 {
@@ -852,7 +855,8 @@ namespace Daqifi.Core.Tests.Device
                 int completionTimeoutMs = 250,
                 CancellationToken cancellationToken = default,
                 Func<CancellationToken, Task>? prepareAsync = null,
-                Func<Task>? finalizeAsync = null)
+                Func<Task>? finalizeAsync = null,
+                bool keepBlankLines = false)
             {
                 try
                 {
--- a/src/Daqifi.Core.Tests/Device/DaqifiDeviceStaleTextLineTests.cs
+++ b/src/Daqifi.Core.Tests/Device/DaqifiDeviceStaleTextLineTests.cs
@@ -365,7 +365,8 @@ public class DaqifiDeviceStaleTextLineTests
             int completionTimeoutMs = 250,
             CancellationToken cancellationToken = default,
             Func<CancellationToken, Task>? prepareAsync = null,
-            Func<Task>? finalizeAsync = null)
+            Func<Task>? finalizeAsync = null,
+            bool keepBlankLines = false)
         {
             try
             {
--- a/src/Daqifi.Core.Tests/Device/DaqifiStreamingDeviceAnalogOutputTests.cs
+++ b/src/Daqifi.Core.Tests/Device/DaqifiStreamingDeviceAnalogOutputTests.cs
@@ -565,7 +565,8 @@ namespace Daqifi.Core.Tests.Device
                 int completionTimeoutMs = 250,
                 CancellationToken cancellationToken = default,
                 Func<CancellationToken, Task>? prepareAsync = null,
-                Func<Task>? finalizeAsync = null)
+                Func<Task>? finalizeAsync = null,
+                bool keepBlankLines = false)
             {
                 cancellationToken.ThrowIfCancellationRequested();
                 setupAction();
--- a/src/Daqifi.Core.Tests/Device/Diagnostics/DeviceDiagnosticsTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Diagnostics/DeviceDiagnosticsTests.cs
@@ -70,22 +70,6 @@ public class DeviceDiagnosticsTests
     }
 
     [Fact]
-    public async Task GetSystemLogAsync_WhenBufferHasEntries_IgnoresTheTerminator()
-    {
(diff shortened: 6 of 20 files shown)


cptkoolbeenz:
## Mutation-tested my own tests, and removed one that could not fail

Rather than assume the three new tests earned their place, I mutated both the pre-change implementation and the fix to see which assertions actually move.

| test | fails on pre-change code? | catches a mutation of the fix? | verdict |
|---|---|---|---|
| `WhenDeviceAnswersNothingAtAll_Throws` | **yes** | — | proves the fix |
| `WhenBufferEmpty_ReturnsEmpty` | no | **yes** | genuine guard |
| `WhenBufferHasEntries_IgnoresTheTerminator` | no | **no** | **vacuous — removed** |

**The removed one asserted something that was never possible.** It checked that a populated dump does not leak its trailing blank line into the parsed entries. But `SystemLogParser.Parse` ignores blank lines, and `ScpiResponseClassifier.IsErrorOnlyResponse` skips them explicitly via `IsNullOrWhiteSpace` — so the terminator could not have become an entry whether the filter ran or not. Deleting the filter line entirely leaves all tests green. That is coverage-shaped nothing, so it is gone rather than left in to pad the count.

**The second one is a guard, not a proof, and I checked it is a real guard.** It passes before and after, so it does not demonstrate the fix — but mutating the throw from *"no line arrived"* to *"every line is blank"* makes it fail. That is exactly the over-eager version of this fix (throwing on a healthy device whose log is simply empty), so the test is worth its place.

**Consequence for the filter itself:** `raw.Where(line => line.Length > 0)` is redundant, and its comment now says so instead of implying otherwise. It stays because it keeps this seam's contract — everything downstream sees the same content lines it saw before `keepBlankLines` existed — true by construction rather than by depending on two other components' internals. But a reader should not mistake it for load-bearing, and now won't.

Revised counts: 4 pre-existing failures unchanged, passing **3531 → 3532**, total 3537 → 3538.

This is the same check I asked the adversarial audit to run as item (d). It found it first.

cptkoolbeenz:
## Adversarial audit: **`gate: PASS`** — the one finding was refuted on a before/after trace

Qodo raised one bug, *"late terminator masks silence"*: a stale blank arriving between the stale-line boundary capture and the send would be counted as this exchange's terminator, so `GetSystemLogAsync` would report an empty log for a silent device.

**The mechanism is real. The regression is not.** The audit traced both `origin/main` and this head (`da8c6a3`) and found identical behaviour in that exact race: on `main`, `EmitEmptyLines = true` already lets such a late blank satisfy `hasReceivedAny`, and the pre-existing `.Where(line => line.Length > 0)` filter reduces it to an empty list — which `SystemLogParser.Parse` and `ThrowIfErrorOnlyResponse` already treat as "no entries, don't throw". Same input, same output, before and after. It is the `staleLineCount` heuristic's documented gap (the #396 comment at the capture site) and it applies to **every** engine caller, not just this one.

## I tried to fix it anyway, then reverted — and the reason is the interesting part

I did not want to decline a bug I could fix, so I implemented a second boundary captured immediately after `setupActionAsync` returns, used **only** to reject blanks (a device cannot emit a terminator for a command it has not been sent; content lines kept the wider boundary so nothing else changed). It built cleanly, and I added two engine-level tests using the existing `ReleaseOnStreamAccessMockTransport` harness.

Then I mutation-tested them, and **both passed with the fix removed.**

The existing `staleLineCount` skip already handles a blank released at consumer-bind — that is *before* the capture. The genuine window is the sub-millisecond interval between the capture and the send completing, and the harness cannot target it: `ReleaseOnStreamAccess(n)` keys off stream accesses, not off that interval.

So I could not demonstrate the fix changed any reachable behaviour. **Shipping an unfalsifiable change plus two tests that cannot fail is worse than a documented limitation**, so it is reverted — this branch is back to exactly the audited head.

Filed as **[#553](https://github.com/daqifi/daqifi-core/issues/553)** with the reproduction gap, why the naive fix is untestable today, and the two decisions a real fix needs (a seam that can reach the window; whether narrowing the content-line boundary risks dropping a fast genuine reply — a bench question, not a reasoning one).

## Final state

| check | result |
|---|---|
| adversarial audit | **PASS**, `treadmill: false`, 1 raw finding / 1 refuted / 0 confirmed |
| `/improve` | 0 suggestions |
| `/agentic_review` | 1 bug, refuted above and tracked as #553 |
| tests | 4 pre-existing failures (unchanged), passing **3531 → 3532** |
| tests that earn their place | 1 fails on pre-change code; 1 guard proven by mutating the fix; 1 removed as vacuous |

Merging.

Issue #536, opened 2026-08-14
feat(diagnostics): memory diagnostics hide the heap-fragmentation fields, so a consumer cannot see how close the device is to the edge


tylerkron:
### What a consumer runs into

`GetMemoryDiagnosticsAsync()` is how an application asks a DAQiFi device how much memory it has left. The most useful thing you can learn from that answer — **whether the heap is fragmented** — is the one thing the API does not have a property for.

On the bench Nq1 (fw 3.7.2) the device reports `HeapFree=7544` out of `HeapTotal=75000`. Read through Core's typed properties that looks like "10% free, a bit tight". The full firmware answer says something much sharper:

```
HeapFree         = 7544
LargestFreeBlock = 7544
SmallestFreeBlock= 7544
HeapFreeBlocks   = 1
```

One free block, and it *is* the entire free heap. So the largest allocation the device can still service is 7.5 KB, and there is no headroom for a second concurrent one. That is a qualitatively different situation from "7.5 KB free spread over twenty blocks", and today a consumer can only find it out by reaching into the untyped bag with a magic string:

```csharp
var largest = m.Values["LargestFreeBlock"];   // no property, no discoverability, no IntelliSense
```

This matters for this library in particular because Core already ships `SdCardEmptyTransferException` for a wedged SD subsystem, and heap exhaustion is the condition behind it. The fields that would let an application *predict* that failure — or explain it after the fact — are exactly the ones without accessors.

### What was measured

Driving a real device (Nq1, fw 3.7.2, USB) through `IDeviceDiagnostics.GetMemoryDiagnosticsAsync()`, the firmware emits **17** `Key=Value` fields. `MemoryDiagnostics` declares **9** typed accessors. The 8 with no property are:

| Field | Value on bench | Why it is worth having |
|---|---|---|
| `LargestFreeBlock` | 7544 | biggest allocation still possible |
| `SmallestFreeBlock` | 7544 | with the above, the fragmentation picture |
| `HeapFreeBlocks` | 1 | free-list fragment count |
| `SdCircularSize` | 512 | SD buffer the card operations run through |
| `SamplePoolBytes` | 22000 | sample-pool sizing |
| `SampleNextFreeBytes` | 2200 | |
| `SampleQueueBytes` | 4480 | |
| `SampleElementBytes` | 20 | |

Parsing itself is correct — all 17 land in `Values` with the right numbers, and every existing typed accessor resolves. This is purely a gap in the typed surface, not a parser bug.

[...]

PR #554, opened 2026-08-20
fix(streaming): report acquisition statistics in the channel's own units

### LINKED ISSUE #534: bug(streaming): acquisition statistics report volts for a channel configured in engineering units
## What a user sees

Configure a channel with engineering units — the feature added in #523, so a pressure transducer reads `12.4 PSI` instead of `2.48 V` — then read an `AcquisitionStatisticsSnapshot`. The min, max and mean it reports for that channel come back **in volts**, not in the unit the channel was configured in, and nothing in the snapshot says which of the two you got.

On the bench (Nq1, fw 3.7.2, analog 1 configured `gain 2.0, offset 10.0, unit "PSI"`, 100 Hz for 15 s, 1,189 samples):

| | `AcquisitionStatistics` reports | every sample's `ScaledValue` says |
|---|---|---|
| min | `0.008545` | `10.017090 PSI` |
| max | `0.010986` | `10.021973 PSI` |
| mean | `0.009570` | `10.019139 PSI` |

Every other number in the same snapshot is correct — sample counts, both sample rates, jitter bounds, latency and out-of-order counts all matched an independently recomputed ground truth exactly. It is only the three value figures.

## Why it can't be worked around

`ChannelAcquisitionStatistics` carries no `Unit` and no gain/offset, so a consumer can't tell a volts reading from an engineering-unit one, and can't convert after the fact — the snapshot has thrown the scaling away. It's also the one place in Core where that loss is unrecoverable: every other path hands the consumer the sample itself, so they can call `.ScaledValue`/`.Unit` for themselves. `AcquisitionStatistics` is the only thing that consumes samples on your behalf and returns just the derived numbers.

Re-applying the scaling by hand is not a safe workaround either, because `ChannelScaling` permits a negative gain (only finiteness is validated) and an inverting transducer swaps the extremes. Same bench, analog 1 at `gain -3.0, offset 5.0`:

- reported `MinValue` 0.008545 V → scaled 4.974365, which is the **maximum** in PSI
- reported `MaxValue` 0.009766 V → scaled 4.970703, which is the **minimum** in PSI

So min and max come back transposed for any inverting scaling. The mean is the only one that survives a manual fix-up, since the transform is linear.

## Root cause

`AcquisitionStatistics` reads `sample.Value` at both entry points — `Device/AcquisitionStatistics.cs:280` (`Record(IChannel, IDataSample)`) and `:518` (the attached `SampleReceived` handler) — rather than `sample.ScaledValue`.

This looks like an ordering accident rather than a decision: `AcquisitionStatistics` landed in #518 at 18:12 on 2026-08-13, and `ChannelScaling` landed in #523 at 22:21 the same day. The statistics class was written four hours before engineering units existed, so it could not have considered them. `AcquisitionStatisticsTests` has zero references to `Scaling` or `ScaledValue`, which is why nothing caught it.

Worth noting the word "scaled" is genuinely overloaded here, so the existing doc comments read either way: `IDataSample.Value`'s own summary calls it "the scaled value of the sample (e.g. volts)" — meaning calibration-scaled — while `ChannelAcquisitionStatistics.MinValue`/`MeanValue` say "the smallest scaled value seen" / "the arithmetic mean of every scaled value". The docs don't settle it; the user-visible consequence above is the argument.

Also relevant: no production code in Core reads `ScaledValue` today — only `StreamFrameDecoderTests` does. So this may be a broader convention question rather than a one-line fix.

## The call to make

Two coherent fixes, and this is a judgement call rather than an obvious one:

1. **Record `ScaledValue`** and add `Unit` to `ChannelAcquisitionStatistics`. Matches what the channel was configured to mean and matches the docs as written. Changes the numbers reported to any existing consumer that has scaling configured.
2. **Keep recording `Value`**, add `Unit` (or the `ChannelScaling` itself) to `ChannelAcquisitionStatistics`, and say plainly in the docs that value statistics are pre-scaling. Non-breaking, and defensible if these figures are meant as a device-health view rather than a measurement view.

Either way `ChannelAcquisitionStatistics` should carry the unit — every analog channel already has one (`CapabilityChannelUnits` gives them an identity scaling labelled `"V"` at connect), so today the snapshot is dropping a label it is already being handed.

## Reproduction

Non-destructive; found during a bench-validation pass on the statistics surface, which had only fake-device tests. Attach an `AcquisitionStatistics` to a connected device, set `((IScaledChannel)channel).Scaling = new ChannelScaling(2.0, 10.0, "PSI")` on one analog channel before streaming, stream, then compare `Snapshot().Channels[n].MeanValue` against the `ScaledValue` of the samples arriving on that channel's `SampleReceived`. Reproduced three times (gain 2.0/offset 10.0, gain -3.0/offset 5.0, and an unscaled control where reported and scaled agree because the scaling is identity).

cptkoolbeenz:
Fixes #534.

## The defect

Configure a channel with engineering units — a pressure transducer reading `12.4 PSI` rather than `2.48 V` — and `AcquisitionStatistics` reported its min, max and mean **in volts**, with nothing in the snapshot saying which of the two you had got.

Both entry points read `sample.Value` instead of `sample.ScaledValue`: `Record(IChannel, IDataSample)` and the attached `SampleReceived` handler.

It looks like an ordering accident rather than a decision — `AcquisitionStatistics` landed in #518 **four hours before** `ChannelScaling` existed (#523, the same day), so it could not have considered engineering units. `AcquisitionStatisticsTests` had zero references to `Scaling` or `ScaledValue`, which is why nothing caught it.

## Why record the scaled value rather than document the volts

The issue laid out both options fairly and called it a judgement call. The deciding evidence is that **a negative gain transposes the extremes**, so recording volts is not merely mislabelled — it cannot be corrected afterwards:

> with `gain -3.0, offset 5.0`, the smallest voltage is the **largest** engineering reading, so a consumer re-applying the scaling by hand gets a `MinValue` that is really the maximum.

`ChannelScaling` permits a negative gain — only finiteness is validated — so that is a supported configuration, not an abuse. Option 2 (keep volts, document it) leaves that consumer with no safe path at all.

It is also the only place in Core where the loss is unrecoverable: every other path hands the consumer the sample so they can read `ScaledValue`/`Unit` themselves. This type consumes the samples on their behalf and returns only the derived numbers.

## `Unit` on the snapshot

`ChannelAcquisitionStatistics` now carries `Unit` — which the issue asked for **either way**, since every analog channel already has one (`CapabilityChannelUnits` gives them an identity scaling labelled `"V"` at connect), so the snapshot was dropping a label it was already being handed.

It follows the same **last-seen-wins** rule as `Name`: a channel's scaling can be reassigned mid-session, and freezing the first unit would label later samples with a unit they were not measured in.

## Behaviour change — worth a release note

A consumer **with** scaling configured sees different numbers. That is the fix. A consumer **without** scaling sees exactly what it saw before, because `ScaledValue` returns `Value` when there is no scaling.

`Unit` is appended with a default, so existing positional construction still compiles.

## Test plan — each test checked for whether it can actually fail

| test | fails on pre-change code? | catches a mutation? | verdict |
|---|---|---|---|
| `ReportsTheEngineeringValueAndItsUnit` | **yes** | — | proves the fix |
| `DoesNotTransposeTheExtremes` | **yes** | — | proves the fix |
| `WithNoScaling_IsUnchangedAndReportsNoUnit` | no | **yes** | control, and a real one |

The third passes before and after, so it does not prove anything on its own — I say so rather than implying all three discriminate. It is still worth keeping: mutating the code to default an unstated unit to `"V"` makes it fail, which is the plausible way this path would regress.

**Baseline-checked, not assumed.** `main` fails **4** tests on this station before any change — all `Device.Discovery.LinuxUsbPortDescriptorProviderTests`, a WSL/Windows runtime artefact unrelated to this work. After this change: the **same 4**, with passing **3532 → 3535**.

No hardware needed; the bench reproduction in the issue is already recorded there.

### DIFF
--- a/src/Daqifi.Core.Tests/Device/AcquisitionStatisticsTests.cs
+++ b/src/Daqifi.Core.Tests/Device/AcquisitionStatisticsTests.cs
@@ -537,20 +537,36 @@ namespace Daqifi.Core.Tests.Device
             var channel = new AnalogChannel(0);
             var sample = Sample(Epoch, 1.0);
 
-            // Warm up: JIT the path and create the channel's accumulator, the one allocation there is.
-            for (var i = 0; i < 100; i++)
+            // Warm up over the SAME count as the measurement, not a token 100.
+            //
+            // Tiered compilation promotes a hot method to tier 1 partway through a long loop, and
+            // that rejit allocates. With a 100-iteration warmup the promotion landed inside the
+            // measured window instead of before it, and the test reported ~7 KB across 10,000
+            // records -- 0.71 bytes per call, which is not a per-call allocation at all (a real one
+            // is >= 24 bytes an object, so >= 240 KB here) but was enough to fail an == 0 assertion.
+            //
+            // That is issue #531: "fails on a full net10 run, passes in isolation". It is not
+            // actually run-order dependent, it is warmup-dependent -- any change that makes
+            // RecordCore larger pushes the promotion later, and #534's scaling-comparison did.
+            // Warming to the measured count puts the promotion firmly before `before` is read.
+            //
+            // The assertion keeps its teeth: a genuine per-call allocation is three orders of
+            // magnitude above the noise this removes.
+            const int iterations = 10_000;
+
+            for (var i = 0; i < iterations; i++)
             {
                 stats.Record(channel, sample);
             }
 
             var before = GC.GetAllocatedBytesForCurrentThread();
-            for (var i = 0; i < 10_000; i++)
+            for (var i = 0; i < iterations; i++)
             {
                 stats.Record(channel, sample);
             }
             var allocated = GC.GetAllocatedBytesForCurrentThread() - before;
 
-            Assert.True(allocated == 0, $"recording 10,000 samples allocated {allocated} bytes; expected none");
+            Assert.True(allocated == 0, $"recording {iterations:N0} samples allocated {allocated} bytes; expected none");
         }
 
         [Fact]
@@ -686,6 +702,93 @@ namespace Daqifi.Core.Tests.Device
             Assert.Equal(1.0, channelStats.MinValue);
             Assert.Equal(3.0, channelStats.MaxValue);
             Assert.Equal(2.0, channelStats.MeanValue);
+        }
+
+        [Fact]
+        public void Record_WhenTheScalingChangesMidWindow_RestartsTheValueFiguresAndSaysSo()
+        {
+            // IScaledChannel supports reassigning a scaling mid-session, and samples keep the one
+            // in force when they were decoded -- so a window can genuinely hold two units. A
+            // minimum in PSI and a maximum in Bar are not the extremes of anything, and reporting
+            // the newest unit over both would put a label on them that is actively false.
+            var clock = new TestClock(Epoch);
+            using var stats = new AcquisitionStatistics(null, clock.Read);
+            var channel = new AnalogChannel(0) { Name = "AI0" };
+            var psi = new ChannelScaling(2.0, 10.0, "PSI");
+            var bar = new ChannelScaling(1.0, 0.0, "Bar");
+
+            foreach (var volts in new[] { 1.0, 3.0 })          // 12, 16 PSI
+            {
+                stats.Record(channel, Sample(clock.Now, volts, psi));
+                clock.Advance(TimeSpan.FromMilliseconds(1));
+            }
+
+            foreach (var volts in new[] { 5.0, 7.0 })          // 5, 7 Bar
+            {
+                stats.Record(channel, Sample(clock.Now, volts, bar));
+                clock.Advance(TimeSpan.FromMilliseconds(1));
+            }
+
+            var channelStats = Assert.Single(stats.Snapshot().Channels);
+
+            // Only the Bar samples describe the value figures.
+            Assert.Equal("Bar", channelStats.Unit);
+            Assert.Equal(5.0, channelStats.MinValue);
+            Assert.Equal(7.0, channelStats.MaxValue);
+            Assert.Equal(6.0, channelStats.MeanValue);
+            Assert.Equal(2, channelStats.ValueSampleCount);
+
+            // Timing and counting are unit-independent and continue across the change.
+            Assert.Equal(4, channelStats.SampleCount);
+        }
+
+        [Fact]
+        public void Record_WhenOnlyTheGainChanges_AlsoRestartsTheValueFigures()
+        {
+            // Compared on the whole scaling, not just the unit label: recalibrating a transducer
+            // leaves the unit reading "PSI" while making the earlier numbers incomparable with the
+            // later ones. Aggregating those is the same error wearing a matching label.
+            var clock = new TestClock(Epoch);
+            using var stats = new AcquisitionStatistics(null, clock.Read);
+            var channel = new AnalogChannel(0) { Name = "AI0" };
+
+            stats.Record(channel, Sample(clock.Now, 1.0, new ChannelScaling(2.0, 0.0, "PSI")));
+            clock.Advance(TimeSpan.FromMilliseconds(1));
+            stats.Record(channel, Sample(clock.Now, 1.0, new ChannelScaling(50.0, 0.0, "PSI")));
+
+            var channelStats = Assert.Single(stats.Snapshot().Channels);
+
+            Assert.Equal("PSI", channelStats.Unit);
+            Assert.Equal(50.0, channelStats.MinValue);
+            Assert.Equal(50.0, channelStats.MaxValue);
+            Assert.Equal(1, channelStats.ValueSampleCount);
+            Assert.Equal(2, channelStats.SampleCount);
+        }
+
+        [Fact]
+        public void Record_WithAStableScaling_CountsEveryValueSample()
+        {
+            // The control for the restart: an unchanging scaling must not trigger it, or every
+            // window would collapse to its last sample.
+            var clock = new TestClock(Epoch);
+            using var stats = new AcquisitionStatistics(null, clock.Read);
+            var channel = new AnalogChannel(0) { Name = "AI0" };
+            var psi = new ChannelScaling(2.0, 10.0, "PSI");
+
+            foreach (var volts in new[] { 1.0, 3.0, 2.0 })
+            {
+                // A NEW but EQUAL scaling instance each time -- ChannelScaling is a record, so
+                // this must compare equal and must not look like a reassignment.
+                stats.Record(channel, Sample(clock.Now, volts, new ChannelScaling(2.0, 10.0, "PSI")));
+                clock.Advance(TimeSpan.FromMilliseconds(1));
+            }
+
+            var channelStats = Assert.Single(stats.Snapshot().Channels);
+            Assert.Equal(3, channelStats.ValueSampleCount);
+            Assert.Equal(3, channelStats.SampleCount);
+            Assert.Equal(12.0, channelStats.MinValue);
+            Assert.Equal(16.0, channelStats.MaxValue);
+            Assert.Equal(psi.Unit, channelStats.Unit);
         }
 
         #endregion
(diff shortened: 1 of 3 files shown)


cptkoolbeenz:
## Audit round 1 and Qodo both landed on the same gap. Fixed — it was a way my own fix made things worse.

Recording the engineering value introduced a failure mode that recording raw volts did **not** have: a window can now hold samples measured in two **different** units.

`IScaledChannel` documents mid-session reassignment as supported — *"set it and every sample decoded from then on carries it"* — and samples keep the scaling in force when decoded. So a channel rescaled PSI → kPa mid-window produced one aggregate spanning both, labelled with only the newest unit. A minimum in PSI and a maximum in kPa are not the extremes of anything, and the label asserted they were kPa.

**Before this PR the window was always volts and therefore at least unit-homogeneous.** So this was a new way to be wrong that my fix created — not a pre-existing gap it inherited. That is the reason it had to be fixed here rather than filed.

The audit recommended resetting the accumulator on a scaling change; that is what landed.

### What restarts, and what deliberately does not

**The value accumulators restart** — `MinValue`, `MaxValue`, `ValueSum` — so the three figures always describe exactly one scaling. **Timing, ordering and `SampleCount` continue untouched**: they are unit-independent, and a rescale is not a gap in the stream.

**Compared on the whole `ChannelScaling`, not just its unit label.** Recalibrating a transducer leaves the unit reading `"PSI"` while making the earlier numbers incomparable with the later ones — aggregating those is the same error wearing a matching label. `ChannelScaling` is a record, so this is structural equality, and a fresh-but-equal instance per sample does not read as a reassignment.

**`ValueSampleCount` is now on the snapshot** so the restart is visible. Without it `MeanValue` would silently describe fewer samples than `SampleCount` reports — an unexplained discrepancy is its own small trap.

### Tests — all three fail without the change

| test | asserts |
|---|---|
| `WhenTheScalingChangesMidWindow_RestartsTheValueFiguresAndSaysSo` | PSI then Bar → `Unit="Bar"`, figures from the Bar samples only, `ValueSampleCount=2` while `SampleCount=4` |
| `WhenOnlyTheGainChanges_AlsoRestartsTheValueFigures` | same unit, different gain → still restarts |
| `WithAStableScaling_CountsEveryValueSample` | a **new but equal** scaling per sample must *not* restart — the control that stops the rule collapsing every window to its last sample |

## Also in this commit: the #531 flake, which my change would have made deterministic

`Record_AfterTheFirstSamplePerChannel_AllocatesNothing` started failing **3/3** in the class run while passing in isolation. I did not assume it was the known flake — I A/B'd it: **production change alone, 30/30 green**; my three new tests added, 32/33. So my production code allocates nothing; the tests perturbed the run.

The failure said **7,128 bytes across 10,000 records**. That is **0.71 bytes per call** — not a per-call allocation at all, since a real one is ≥24 bytes an object and so ≥240 KB here. It is a one-off inside the measured window: **tiered compilation** promoting a now-larger `RecordCore` to tier-1, which the 100-iteration warmup no longer outran.

So [#531](https://github.com/daqifi/daqifi-core/issues/531)'s *"fails on a full net10 run, passes in isolation"* is **not run-order dependent — it is warmup-dependent**. Any change that grows `RecordCore` pushes the promotion later. Warming over the same count as the measurement puts it firmly before measurement begins: **3/3 green** where it was 3/3 red. The assertion keeps its teeth, since a genuine per-call allocation is three orders of magnitude above the noise removed.

Baseline unchanged: same 4 pre-existing failures, passing 3532 → **3538**.

Round 2 of the audit is running.

cptkoolbeenz:
## Qodo pass 2: mixed-unit bug **✓ Resolved**. One new finding — real mechanism, wrong place to fix it.

### New: *"Overflow fallback mislabels volts"*

**The mechanism is real**, and I verified it in source rather than taking it on faith. `ChannelScaling.Apply` (`ChannelScaling.cs:101`):

```csharp
var scaled = value * Gain + Offset;
return double.IsFinite(scaled) ? scaled : value;   // <- the original, in the original unit
```

So on overflow `ScaledValue` is **volts** while `Unit` still says `"PSI"`, and nothing downstream can tell — the return type is a bare `double` with no signal that the fallback fired.

**But this PR cannot fix it, and did not cause it.** Given a bare double, `AcquisitionStatistics` has no way to know the conversion was skipped. The fix has to live where the information exists — a `TryApply`, or a documented contract change on `Apply`. Filed as **[#555](https://github.com/daqifi/daqifi-core/issues/555)** with three options and a recommendation.

**Reachability, measured rather than asserted.** Overflow needs `value * Gain + Offset` to leave double range, and both coefficients are already validated finite:

| reading | gain needed to overflow |
|--:|--:|
| 1.0 V | ~1.80e308 |
| 10.0 V | ~1.80e307 |
| 4095 | ~4.39e304 |

Not a transducer configuration anyone types — an adversarial input. And the **non-finite-input** case is unaffected: `NaN`/`±∞` in gives a non-finite result out, which is visibly wrong rather than silently mislabelled. That is the fallback working as designed.

**One thing worth saying plainly rather than burying:** before this PR, *no production code in Core read `ScaledValue`* (per #534 — only `StreamFrameDecoderTests` did). So this PR is the **first production consumer to expose** the hazard, even though it did not introduce it. That is why it is filed rather than shrugged off.

### The remaining open suggestion is the same item

`Prevent mixed-unit statistical aggregation` (importance 9) is the pass-1 finding, which is **fixed** in `2629b0f` — Qodo's own bug counter marks it ✓ Resolved. It stays listed because `/agentic_review` does not un-list a finding once raised.

### State

| gate | result |
|---|---|
| `/improve` | 1 open — the already-fixed mixed-unit item |
| `/agentic_review` | 2 bugs: 1 ✓ Resolved, 1 declined → [#555](https://github.com/daqifi/daqifi-core/issues/555) |
| adversarial audit round 2 | running against `2629b0f` |
| tests | 4 pre-existing failures, passing 3532 → **3538** |

cptkoolbeenz:
## Audit round 2: **`ready_to_merge`**, `treadmill: true`

The round-1 mixed-unit defect is resolved — the arbiter confirms the structural `ChannelScaling` comparison, the mid-window accumulator restart, `ValueSampleCount`, and the three tests close it.

One finding remains, dispositioned **`defer_ticket`** at low severity, and it is a genuine one worth stating rather than glossing:

### Appending positional record fields is binary-breaking

`ChannelAcquisitionStatistics` goes from **14 to 16** positional parameters. Optional defaults preserve **source** compatibility — anyone who rebuilds is fine — but **not binary** compatibility, because C# emits one `.ctor`/`Deconstruct` at the new arity and resolves defaults at the *caller's* compile time. A consumer compiled against v1.6.0's 14-parameter shape, run against the new DLL without recompiling, throws `MissingMethodException`.

The auditor **reproduced this empirically** with a minimal record rather than reasoning about it, which is why I am not arguing with the mechanism.

**Why it does not block, and why that is not special pleading:** it fails **loud** — a hard exception at the call site — where round 1's mixed-unit bug failed **silently** into a wrong-but-plausible number. That asymmetry is exactly why that one was fixed here and this one is deferred. The break is also narrow (property reads, `with`, `Equals`, `GetHashCode`, `ToString` all unaffected; only positional construction and full-arity `Deconstruct`), and this type is a read-only snapshot consumers overwhelmingly read rather than build.

Filed as **[#557](https://github.com/daqifi/daqifi-core/issues/557)** — a release-process decision, not a code change: accept-and-changelog, add a compatibility overload, or state a policy. My preference is accept-and-changelog **plus** writing the policy down once, because this type will keep gaining fields as the firmware reports more (#536 is adding eight accessors to a sibling type right now), so a shim per addition does not scale.

Reworking the record shape this late in the PR would be scope creep on a fix that is otherwise done, and the arbiter flagged `treadmill: true`.

### Qodo

`review_bugs=2`: bug 1 (mixed units) **✓ Resolved**; bug 2 (`ChannelScaling.Apply`'s overflow fallback) declined with measured reachability and filed as **[#555](https://github.com/daqifi/daqifi-core/issues/555)**. The single open suggestion is the already-fixed mixed-unit item, which `/agentic_review` re-lists rather than un-lists.

### Final state

| check | result |
|---|---|
| adversarial audit round 2 | `ready_to_merge`, 1 finding deferred → #557 |
| Qodo bugs | 1 resolved, 1 declined → #555 |
| tests | 4 pre-existing failures (unchanged), passing 3532 → **3538** |
| tests that earn their place | 5 fail on pre-change code; 2 are mutation-proved controls |
| bonus | [#531](https://github.com/daqifi/daqifi-core/issues/531) root-caused as warmup-dependent, not run-order dependent, and fixed |

Merging.

PR #556, opened 2026-08-20
feat(diagnostics): expose the heap-fragmentation and pool fields as properties

### LINKED ISSUE #536: feat(diagnostics): memory diagnostics hide the heap-fragmentation fields, so a consumer cannot see how close the device is to the edge

cptkoolbeenz:
Fixes #536.

`GetMemoryDiagnosticsAsync` is how an application asks a device how much memory it has left — and the sharpest thing the firmware tells you, **how the free heap is broken up**, had no property. A consumer could only reach it through the untyped bag with a magic string:

```csharp
var largest = m.Values["LargestFreeBlock"];   // no property, no discoverability, no IntelliSense
```

The firmware sends **17** fields; `MemoryDiagnostics` declared **9** accessors. The missing eight now have them:

`LargestFreeBlock` · `SmallestFreeBlock` · `HeapFreeBlocks` · `SdCircularSize` · `SamplePoolBytes` · `SampleNextFreeBytes` · `SampleQueueBytes` · `SampleElementBytes`

Purely additive — same `ulong?` / `GetValue` shape as the existing nine, still `null` when the firmware of the day omits a field, so the *"field set varies by firmware version"* contract the type already documents is unchanged. **Parsing was never wrong**: all 17 already landed in `Values` with the right numbers. This was a gap in the typed surface only.

`LargestFreeBlock` is the one that earns its place: it, not `HeapFree`, is the largest allocation the device can still service. 7.5 KB free across twenty blocks and 7.5 KB free in one block report the same `HeapFree` and have completely different prospects.

## One correction to the issue

The issue offered an optional helper, `IsHeapFragmented`, defined as `LargestFreeBlock == HeapFree && HeapFreeBlocks == 1`.

**That predicate is backwards.** One free block that *is* the entire free heap is the **least** fragmented state possible — a single unbroken run. What the bench device was showing is not fragmentation but **near-exhaustion without fragmentation**, which is a different diagnosis with a different remedy: fragmentation says "the bytes are there but scattered, compact or restart", while this says "the bytes are simply gone".

So I have shipped the accessors and left the helper out — which the issue explicitly allowed for (*"Happy to leave that out if it feels like over-reach; the accessors are the part that matters"*) — and put the distinction in the doc comment on `HeapFreeBlocks` instead:

> A **high** count means fragmentation … A count of `1` is the opposite — the free heap is one contiguous run, the least fragmented state there is. A device can therefore be unfragmented and nearly exhausted at the same time.

A helper that named that state "fragmented" would send someone looking for the wrong problem, which is worse than making them read two numbers.

## Test plan

Two tests, using the **real 17-field response** measured on an Nq1 running 3.7.2 (from the issue), so the fixture is what a device actually says rather than one I invented:

| test | asserts |
|---|---|
| `ExposesTheFragmentationAndPoolFieldsAsProperties` | all eight resolve; `HeapFree == LargestFreeBlock`, the reading the fields exist to make possible |
| `WhenFirmwareOmitsTheNewFields_TheyReadNull` | an older device gives `null`, not `0` |

The second is **mutation-proved**: making an accessor return `0` instead of `null` fails it. That matters more than it looks — zero would read as *"no free block at all"*, the alarming opposite of *"did not say"*.

The first cannot compile without the accessors, so it is discriminating by construction.

**Baseline-checked:** `main` fails 4 tests on this station before any change (`Device.Discovery.LinuxUsbPortDescriptorProviderTests`, a WSL/Windows runtime artefact). After: the **same 4**, passing 3532 → 3534.

No hardware needed; the bench measurement is already recorded in the issue.

### DIFF
--- a/src/Daqifi.Core.Tests/Device/Diagnostics/MemoryDiagnosticsParserTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Diagnostics/MemoryDiagnosticsParserTests.cs
@@ -40,6 +40,78 @@ public class MemoryDiagnosticsParserTests
         Assert.Equal(11, mem.Values.Count);
     }
 
+    // The full 17-field answer a real Nq1 on firmware 3.7.2 gives, values as measured on the
+    // bench (issue #536). The device was nearly out of heap, which is what makes it a useful
+    // fixture: HeapFree looks merely tight until you see that the whole of it is one block.
+    private static readonly string[] BenchResponse =
+    {
+        "HeapTotal=75000",
+        "HeapFree=7544",
+        "HeapUsed=67456",
+        "HeapMinEverFree=7544",
+        "LargestFreeBlock=7544",
+        "SmallestFreeBlock=7544",
+        "HeapFreeBlocks=1",
+        "CoherentPoolTotal=32768",
+        "CoherentPoolFree=16384",
+        "SdCircularSize=512",
+        "SamplePoolCount=1100",
+        "SamplePoolInUse=4",
+        "SamplePoolMaxUsed=12",
+        "SamplePoolBytes=22000",
+        "SampleNextFreeBytes=2200",
+        "SampleQueueBytes=4480",
+        "SampleElementBytes=20",
+    };
+
+    [Fact]
+    public void TryParse_ExposesTheFragmentationAndPoolFieldsAsProperties()
+    {
+        // These eight parsed correctly all along -- they landed in Values with the right numbers.
+        // What was missing was any way to reach them except by magic string, on the one question
+        // a consumer actually asks this API: how close to the edge is the device?
+        var ok = MemoryDiagnosticsParser.TryParse(BenchResponse, out var mem);
+
+        Assert.True(ok);
+        Assert.NotNull(mem);
+
+        Assert.Equal(7544UL, mem!.LargestFreeBlock);
+        Assert.Equal(7544UL, mem.SmallestFreeBlock);
+        Assert.Equal(1UL, mem.HeapFreeBlocks);
+        Assert.Equal(512UL, mem.SdCircularSize);
+        Assert.Equal(22000UL, mem.SamplePoolBytes);
+        Assert.Equal(2200UL, mem.SampleNextFreeBytes);
+        Assert.Equal(4480UL, mem.SampleQueueBytes);
+        Assert.Equal(20UL, mem.SampleElementBytes);
+
+        // The reading the fields exist to make possible: the largest allocation still serviceable
+        // is the entire free heap, because it is one unbroken run. Unfragmented -- and nearly out.
+        Assert.Equal(mem.HeapFree, mem.LargestFreeBlock);
+        Assert.Equal(17, mem.Values.Count);
+    }
+
+    [Fact]
+    public void TryParse_WhenFirmwareOmitsTheNewFields_TheyReadNull()
+    {
+        // The field set varies by firmware version, which the type already documents. An older
+        // device that does not send these must give null rather than zero -- zero would read as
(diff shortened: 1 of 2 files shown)


cptkoolbeenz:
## Both gates clean on the first pass

| gate | result |
|---|---|
| `/improve` | 0 suggestions |
| `/agentic_review` | **0 bugs**, converged |
| adversarial audit | **PASS**, `clean` — 0 raw findings |

The audit was pointed specifically at the things most likely to be wrong here — whether the eight field-name literals match what the firmware actually emits (a typo would make an accessor return `null` forever, which is exactly the quiet failure this type is prone to), whether `ulong?` is right for each, and whether the `HeapFreeBlocks` doc claim is correct that a count of 1 is the *least* fragmented state. Nothing came back.

Merging.

PR #558, opened 2026-08-20
fix(device): give health telemetry a way to refresh, and stop the docs claiming it is live

### LINKED ISSUE #535: bug(device): health telemetry is documented as live but is frozen at connect, and board temperature is never populated by any supported firmware
**Type**: Bug
**Priority**: Medium

## What a user sees

`device.Metadata.Health` is documented as live telemetry — battery charge, board temperature, power state — that keeps up with the device. It is not. It is a snapshot taken once, when you connect, and it never changes again for the life of the connection.

Build a UI that shows "Battery: 87%" and it will still say 87% an hour later, on a device that has since run flat. Ask for the board temperature and you get `null`, always, on every firmware DAQiFi ships.

Both halves are confirmed against the bench Nq1 on fw 3.7.2, and both are traceable to firmware source.

## Confirmed on hardware

A 65-second live connection with three analog channels streaming at 100 Hz, instrumented on `StatusMessageReceived` / `StreamMessageReceived` / `MessageReceived`:

| window | duration | stream frames | status frames Core received |
|---|---|---|---|
| idle, nothing sent | 40 s | 0 | **0** |
| streaming @ 100 Hz | 20 s | 1,586 | **0** |
| after one explicit `GetDeviceInfo` | 2.5 s | — | 1 |

1,587 inbound protobuf messages; exactly **one** carried health, and only because the harness asked for it. `Health` read `Battery=100 Temp=null Pwr=1 DevStatus=3` at connect and was byte-identical at every checkpoint through the whole run.

A separate run sending `GetDeviceInfo` five times at 2-second intervals got exactly five status frames — one per request, none in between. The device answers when asked and is otherwise silent.

## Why — two independent causes

**1. Nothing ever asks again.** The only `GetDeviceInfo` sends in `DaqifiDevice` are inside `WaitForChannelsPopulatedAsync` (`DaqifiDevice.cs:3814`, `:3860`), which stops the moment channels populate during `InitializeAsync`. After that Core never requests another status message, and there is no public `RefreshStatus`-style API to do it yourself — the only way in is `device.Send(ScpiMessageProducer.GetDeviceInfo)`, which works but is undocumented as the refresh path.

**2. The firmware emits no periodic status frames on a connected link.** `streaming.c:2144-2153` rebuilds the field list per frame from exactly four tags — `msg_time_stamp`, `analog_in_data`, `digital_data`, `digital_port_dir`. No health fields. `DeviceHealth.cs:6` states values update "including the periodic ones emitted during streaming"; there are none. (Health fields *are* in the periodic **WiFi UDP discovery broadcast** — `fields_discovery`, `app_freertos.c:166` — but that is the discovery channel, which a connected device never consumes. The original ticket #335 hedged this correctly as "if present"; the hedge was dropped when the doc was written.)

## Board temperature is dead on every supported firmware

```c
case DaqifiOutMessage_temp_status_tag:
    break;                    // NanoPB_Encoder.c — no assignment
```

The encoder case is empty, so `temp_status` is never set, stays 0, and nanopb's proto3 encoding omits it. Checked at **v3.5.0, v3.6.0, v3.7.0, v3.7.1, v3.7.2** — empty in all of them, i.e. across the entire range Core supports (ADR-0001). Bench confirms: `TempStatus == 0` in 9/9 status frames, `BoardTemperatureCelsius == null` throughout.

`BoardTemperatureCelsius` therefore cannot be non-null against real hardware, but its doc comment reads as an ordinary available reading.

## Scope

`DeviceHealth` currently has **zero consumers** — nothing in `Daqifi.Core`, `Daqifi.Mcp`, or the example app reads it. So nothing is visibly broken today; the cost is that #335's stated value ("unlocks battery/thermal monitoring for every consumer", "the MCP server could answer 'what's the battery at?'") is not actually available, and the next consumer to wire it up inherits a frozen number with no indication it is stale.

## Suggested fix

Open question for you, which is why this is filed rather than fixed:

1. **Correct the docs and add a refresh.** Drop the "periodic ones during streaming" claim, note on `BoardTemperatureCelsius` that no firmware >= 3.5.0 populates it, and add a public `RefreshDeviceStatusAsync(CancellationToken)` that sends `GetDeviceInfo` and completes on the next status message. Additive, non-breaking. Consumers then poll at whatever cadence they want.
2. **Docs only.** Fix the two claims and document `Send(ScpiMessageProducer.GetDeviceInfo)` as the refresh path.

Option 1 is the smaller surprise for consumers; option 2 is smaller. Worth noting either way that a periodic self-poll inside Core would contend for the operation lock against streaming and SD work, so refresh cadence probably belongs to the caller, not to Core.

## Acceptance criteria

- [ ] `DeviceHealth`'s "periodic ones emitted during streaming" claim no longer asserts something no firmware does
- [ ] `BoardTemperatureCelsius` documents that shipping firmware (>= 3.5.0) never populates it
- [ ] A supported way to refresh health telemetry exists and is documented
- [ ] Test covering the refresh path against a status message

## Evidence

Bench: Nq1 fw 3.7.2, `/dev/cu.usbmodem1101`, serial, non-destructive (connect / stream / disconnect only). Two runs, consistent. Firmware source read at tags v3.5.0 through v3.7.2.

cptkoolbeenz:
Fixes #535.

`device.Metadata.Health` is documented as live telemetry. It is a snapshot taken once at connect that **never changes again** for the life of the connection. Build a UI on it and it shows "Battery: 87%" an hour later on a device that has since run flat.

## Two independent causes, both confirmed on hardware

**1. Nothing ever asks again.** The only `GetDeviceInfo` sends are inside `WaitForChannelsPopulatedAsync`, which stops the moment channels populate during `InitializeAsync`. There was no public API to ask afterwards.

**2. The firmware sends no periodic status frames on a connected link.** Each stream frame is built from four tags and none of them is health, so a connected device is silent unless asked. From the issue's bench run (Nq1, fw 3.7.2):

| window | duration | stream frames | status frames |
|---|---|--:|--:|
| idle | 40 s | 0 | **0** |
| streaming @ 100 Hz | 20 s | 1,586 | **0** |
| after one explicit `GetDeviceInfo` | 2.5 s | — | 1 |

1,587 inbound messages; exactly **one** carried health, and only because it was requested.

## The fix

`RefreshDeviceStatusAsync(TimeSpan?, CancellationToken)` — sends `GetDeviceInfo`, waits for the next status message, returns once it has been applied. Additive, non-breaking.

It **subscribes before sending**, so a fast reply cannot slip through the gap between `Send()` and the subscription — the same ordering `WaitForChannelsPopulatedAsync` already relies on. It **unsubscribes in a `finally`**, so a caller polling in a loop does not accumulate one handler per call for the life of the connection. Default timeout 5 s: a device that is answering at all replies in milliseconds, so this is a hang detector, not a prediction.

**Cadence stays with the caller, deliberately.** The issue flagged this and it is right — a self-poll inside Core would contend for the operation lock against streaming and SD work, and how often "current" needs to be is a decision only the application can make.

## Docs corrected — both claims were wrong

`DeviceHealth` said its values update *"including the periodic ones emitted during streaming"*. There are none. Worth noting the original ticket #335 hedged this correctly as *"if present"*; **the hedge was dropped when the doc was written.**

`BoardTemperatureCelsius` now says what it is: **no shipping firmware populates it.** The encoder's `temp_status` case is empty, so the field is never assigned, stays 0, and proto3 omits it — checked at v3.5.0, v3.6.0, v3.7.0, v3.7.1 and v3.7.2, i.e. the entire range Core supports. Against real hardware it cannot be anything but `null`. Kept, because the wire field exists and a future firmware may fill it.

## Which option this took, and why

The issue offered docs-only or docs-plus-refresh and left it open. The **acceptance criteria settle it** — they ask for *"a supported way to refresh health telemetry"* and *"a test covering the refresh path"*, which docs-only cannot satisfy. It is also the smaller surprise: the alternative documents `Send(ScpiMessageProducer.GetDeviceInfo)` as the refresh path, which works but leaves every consumer to reinvent the wait, the timeout and the unsubscribe.

## Test plan

| test | asserts |
|---|---|
| `CompletesAndAppliesTheReply_WhenAStatusArrives` | it actually sends `SYSInfoPB`, and `Metadata.Health` carries the reply on return |
| `WhenTheDeviceStaysSilent_Throws` | `TimeoutException` rather than a hang — the realistic failure, since a lost request produces silence, not an error frame |
| `WhenNotConnected_Throws` | `DeviceNotConnectedException` |
| `UnsubscribesAfterCompleting` | three refreshes leave nothing subscribed; a later unsolicited status still applies and faults nothing |

All four fail without the new method — it does not exist to be called. The device double now captures `Send` rather than transmitting (it has no transport), which also lets the first test assert *which* command went out rather than merely that the call did not throw.

**Baseline-checked:** `main` fails 4 tests on this station before any change (`Device.Discovery.LinuxUsbPortDescriptorProviderTests`, a WSL/Windows runtime artefact). After: the **same 4**, passing 3538 → 3542.

No hardware needed; the bench evidence is recorded in the issue.

### DIFF
--- a/src/Daqifi.Core.Tests/Device/DaqifiDeviceTests.cs
+++ b/src/Daqifi.Core.Tests/Device/DaqifiDeviceTests.cs
@@ -197,6 +197,41 @@ namespace Daqifi.Core.Tests.Device
         }
 
         [Fact]
+        public async Task RefreshDeviceStatusAsync_ConcurrentCalls_AreNotCompletedByEachOthersReply()
+        {
+            // Both callers subscribe to the same multicast StatusMessageReceived event, so without
+            // serialization ONE incoming frame -- which answers only one of the two requests --
+            // completes both, and the loser returns a success not tied to its own request.
+            var device = new TestableDaqifiDevice("TestDevice");
+            device.Connect();
+
+            var first = device.RefreshDeviceStatusAsync(TimeSpan.FromSeconds(5));
+
+            // The second call must be queued behind the first, not racing it: at this point only
+            // one request can have gone out.
+            var second = device.RefreshDeviceStatusAsync(TimeSpan.FromSeconds(5));
+
+            await Task.Delay(50);
+            Assert.Single(device.SentCommands);
+            Assert.False(second.IsCompleted);
+
+            // One reply satisfies the first caller only.
+            device.InvokeStatusMessage(new DaqifiOutMessage { BattStatus = 11 });
+            await first;
+
+            await Task.Delay(50);
+            Assert.False(second.IsCompleted);
+
+            // The second caller's own request goes out once it holds the gate, and its own reply
+            // completes it.
+            Assert.Equal(2, device.SentCommands.Count);
+            device.InvokeStatusMessage(new DaqifiOutMessage { BattStatus = 22 });
+            await second;
+
+            Assert.Equal(22, device.Metadata.Health.BatteryPercent);
+        }
+
+        [Fact]
         public async Task RefreshDeviceStatusAsync_UnsubscribesAfterCompleting()
         {
             // The handler is added per call, so a caller polling in a loop would otherwise
--- a/src/Daqifi.Core.Tests/Device/DaqifiDeviceTests.cs
+++ b/src/Daqifi.Core.Tests/Device/DaqifiDeviceTests.cs
@@ -197,6 +197,27 @@ namespace Daqifi.Core.Tests.Device
         }
 
         [Fact]
+        public async Task RefreshDeviceStatusAsync_IsNotBlockedByAThrowingStatusSubscriber()
+        {
+            // .NET stops invoking a multicast delegate at the first exception, and
+            // RaiseClassifiedEvent wraps the WHOLE list in one try/catch -- so a consumer that
+            // subscribed earlier and throws would have prevented the refresh from ever seeing its
+            // own reply. It would then time out while the status it asked for had already been
+            // applied to Metadata, which is the most confusing failure this API could produce.
+            var device = new TestableDaqifiDevice("TestDevice");
+            device.Connect();
+
+            device.StatusMessageReceived += _ => throw new InvalidOperationException("boom");
+
+            var refresh = device.RefreshDeviceStatusAsync(TimeSpan.FromSeconds(2));
+            device.InvokeStatusMessage(new DaqifiOutMessage { BattStatus = 55 });
+
+            await refresh;
+
+            Assert.Equal(55, device.Metadata.Health.BatteryPercent);
+        }
+
+        [Fact]
         public async Task RefreshDeviceStatusAsync_ConcurrentCalls_AreNotCompletedByEachOthersReply()
         {
             // Both callers subscribe to the same multicast StatusMessageReceived event, so without
--- a/src/Daqifi.Core.Tests/Device/DaqifiDeviceTests.cs
+++ b/src/Daqifi.Core.Tests/Device/DaqifiDeviceTests.cs
@@ -152,10 +152,93 @@ namespace Daqifi.Core.Tests.Device
             Assert.NotNull(raised);
         }
 
+        [Fact]
+        public async Task RefreshDeviceStatusAsync_CompletesAndAppliesTheReply_WhenAStatusArrives()
+        {
+            // The point of the API: health is not live -- the device answers when asked and is
+            // otherwise silent -- so there has to be a supported way to ask (issue #535).
+            var device = new TestableDaqifiDevice("TestDevice");
+            device.Connect();
+
+            var refresh = device.RefreshDeviceStatusAsync(TimeSpan.FromSeconds(5));
+
+            // The reply, delivered the way the consumer loop delivers one.
+            device.InvokeStatusMessage(new DaqifiOutMessage { BattStatus = 42 });
+
+            await refresh;
+
+            // It must actually ask -- the whole point is that the device says nothing unprompted.
+            Assert.Contains(device.SentCommands, c => c.Contains("SYSInfoPB", StringComparison.Ordinal));
+
+            // Metadata is updated before the event is raised, so it is already applied here.
+            Assert.Equal(42, device.Metadata.Health.BatteryPercent);
+        }
+
+        [Fact]
+        public async Task RefreshDeviceStatusAsync_WhenTheDeviceStaysSilent_Throws()
+        {
+            // A silent device must surface as a timeout rather than hanging the caller. This is
+            // the realistic failure: the device sends nothing at all unless asked, so a lost
+            // request produces silence, not an error frame.
+            var device = new TestableDaqifiDevice("TestDevice");
+            device.Connect();
+
+            await Assert.ThrowsAsync<TimeoutException>(
+                () => device.RefreshDeviceStatusAsync(TimeSpan.FromMilliseconds(150)));
+        }
+
+        [Fact]
+        public async Task RefreshDeviceStatusAsync_WhenNotConnected_Throws()
+        {
+            var device = new TestableDaqifiDevice("TestDevice");
+
+            await Assert.ThrowsAsync<DeviceNotConnectedException>(
+                () => device.RefreshDeviceStatusAsync(TimeSpan.FromSeconds(1)));
+        }
+
+        [Fact]
+        public async Task RefreshDeviceStatusAsync_UnsubscribesAfterCompleting()
+        {
+            // The handler is added per call, so a caller polling in a loop would otherwise
+            // accumulate one subscription per refresh for the life of the connection.
+            var device = new TestableDaqifiDevice("TestDevice");
+            device.Connect();
+
+            for (var i = 0; i < 3; i++)
+            {
+                var refresh = device.RefreshDeviceStatusAsync(TimeSpan.FromSeconds(5));
+                device.InvokeStatusMessage(new DaqifiOutMessage { BattStatus = 7 });
+                await refresh;
+            }
+
+            // A status arriving with no refresh outstanding must not fault anything, and the
+            // device must still be usable afterwards.
+            var ex = Record.Exception(() => device.InvokeStatusMessage(new DaqifiOutMessage { BattStatus = 9 }));
+
+            Assert.Null(ex);
+            Assert.Equal(9, device.Metadata.Health.BatteryPercent);
+        }
+
         private sealed class TestableDaqifiDevice : DaqifiDevice
         {
             public TestableDaqifiDevice(string name, IPAddress? ipAddress = null) : base(name, ipAddress)
             {
+            }
+
+            /// <summary>Commands the device was asked to send, in order.</summary>
+            public List<string> SentCommands { get; } = new();
+
+            /// <summary>
+            /// Captures instead of transmitting: this double has no transport, and the base
+            /// implementation throws without one. Recording the text also lets a test assert
+            /// WHICH command was sent, not merely that the call did not throw.
+            /// </summary>
+            public override void Send<T>(IOutboundMessage<T> message)
+            {
+                if (message is IOutboundMessage<string> text)
+                {
+                    SentCommands.Add(text.Data);
+                }
             }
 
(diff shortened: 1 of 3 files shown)


cptkoolbeenz:
## Audit round 1: `keep_fixing` — one confirmed finding **and one unsound refutation**. Both addressed.

The interesting outcome is the second one: the arbiter rejected its own skeptic's refutation for only testing the single-refresh case, and sent back the exact scenario my audit brief had asked about as check (a). Qodo independently raised the same thing as a suggestion. Two reviewers converging on it is a fair signal that my reasoning was answering the wrong question.

### Concurrent refreshes could complete each other

Two calls to `RefreshDeviceStatusAsync` both subscribed to the same multicast `StatusMessageReceived` event, so **one** incoming frame — which answers only one of the two outstanding requests — completed **both** `TaskCompletionSource`s. The loser returned success off a reply that was never its own.

I want to be precise about the scope, because it is narrower than it sounds and I nearly used that to dismiss it: **no caller ever got wrong data.** The reply updates the shared `Metadata` either way, so both saw a genuinely fresh reading. What was wrong is the **promise** — a method whose whole purpose is "ask the device now and tell me when it answered" should not report an answer to a question it did not ask. And a caller retrying after its own short timeout while a first attempt is still in flight is ordinary usage, not exotic.

A `SemaphoreSlim(1,1)` now serializes them: each call sends its own request and completes on its own reply.

**Mutation-proved.** `ConcurrentCalls_AreNotCompletedByEachOthersReply` asserts the queued caller has not sent yet, is not completed by the first reply, then sends and completes on its own. Widening the semaphore to `SemaphoreSlim(int.MaxValue, int.MaxValue)` makes it fail.

### The timeout could expire before the request was ever sent

Confirmed, and real: `Send` is fire-and-forget, and `OperationSerializer.TryDeferSend` parks it while an exclusive operation holds the device — an SD download can hold it for many minutes. A short timeout taken during one elapses before the device is asked, and the caller gets a `TimeoutException` that reads like a silent device.

The arbiter offered *(a)* document that the deadline covers queue-wait, or *(b)* start the clock only once the send is confirmed transmitted. **I took (a), in-PR rather than as a follow-up ticket.** Behaviour is unchanged, because the caller's deadline bounding *total time to answer* is the defensible reading and it fails **loudly** rather than silently — but it was undocumented, and that is the part that would have cost someone an afternoon. The parameter doc now states the deadline covers waiting behind another refresh and waiting for the send to leave the queue, and names the SD-download case specifically.

One linked `CancellationTokenSource` now covers the gate wait, the send and the reply wait, so there is a single honest deadline rather than one that silently started before part of the work.

### Separately: the baseline is not what I have been calling it

While confirming #531's flake was gone (it is — `AllocatesNothing` green across 9 runs since #554), a **different** test failed in 1 of 4 full runs on unmodified `main`: `LiveSampleStreamTests.ConcurrentEnumerations_AllEndOnTheSameDrop`. Filed as **[#559](https://github.com/daqifi/daqifi-core/issues/559)**.

That matters for how I have been reporting these PRs: "the baseline is 4 failures" is really "4 deterministic plus at least one intermittent". I spent four runs establishing the fifth one was not mine, which is the cost the ticket exists to save the next person.

Round 2 of the audit is running.

cptkoolbeenz:
## Round 2 audit: **PASS / `ready_to_merge`**. Qodo pass 3: three ✓ Resolved, one new — documented.

### Audit round 2 (head `c52c684`) — clean

0 confirmed findings. It also **soundly refuted** one item, and the arbiter checked the refutation rather than taking it: `RefreshDeviceStatusAsync` not being on `IStreamingDevice` is not an asymmetry this PR introduced — `DaqifiDeviceFactory.ConnectTcpAsync`/`ConnectSerialAsync` return the **concrete** `DaqifiStreamingDevice`, the documented Quick Start uses `var`, and every other concrete-only member follows the same pattern — including `InitializeAsync`, which is what populates `Metadata.Health` in the first place.

### Qodo pass 3 (head `cb1c32d`)

`Subscriber can block refresh`, `Refresh bypasses operation serialization` and `Boundary reply becomes timeout` are all **✓ Resolved**. One new finding, and it is a real one:

**A late reply from a timed-out refresh can satisfy the next one.** Same shape as the concurrent case, but sequential — and reachable on a merely *slow* device rather than a silent one.

**I have documented it rather than fixed it, because there is nothing to fix it with.** A status frame carries nothing identifying which request it answers. I checked `MsgTimeStamp` as a possible correlation key: nothing in the device layer reads it for status, and the device-side counter wraps roughly every 102 seconds, so it is not usable as one.

The obvious heuristic — skip one frame after a timeout — **makes the commoner case worse**: if the earlier reply never arrives at all, the next refresh discards its *own* reply and times out too. Trading a rare mild staleness for a common hard failure is a bad trade, so I did not make it.

Worth stating what the caller actually gets, because it is far less alarming than "wrong answer": a **genuine device reading**, taken before their call, bounded by the previous refresh's timeout, with their own reply landing moments later and updating `Metadata.Health` again.

This is the third finding arising from one underlying fact — the protocol has no request/response correlation for status — so the remark names that root cause once rather than cataloguing another symptom. The concurrent case had the same shape for a different reason and **is** prevented outright, by serializing.

### On the last commit and the audit

`c9cabd9` is **documentation only** — one `<para>` in an XML doc comment, no executable change, tests unchanged at 4 pre-existing failures / 3544 passing. Audit round 3 is running against `cb1c32d`, the last commit with any behavioural content. I am not spinning a fourth adversarial round over a doc paragraph; that is the treadmill the arbiter warned about two PRs ago, and there is no code in it for an adversary to find a bug in.

cptkoolbeenz:
## Audit round 3: **PASS, clean, 0 findings.** Merging.

Three adversarial rounds and three Qodo passes. Every finding across both was real; none were declined on the merits except the one the protocol makes unfixable, and that is documented rather than waved off.

### Final state of the two open counters

**`suggestions_unresolved=1` — `Serialize concurrent status refreshes` (importance 5).** This is **implemented**, not outstanding; Qodo simply has not flipped it. Verified in the diff:

```
511:  private readonly SemaphoreSlim _statusRefreshGate = new(1, 1);
3390:     await _statusRefreshGate.WaitAsync(timeoutCts.Token)
3443:     _statusRefreshGate.Release();
```

**`review_bugs=1` — `Late reply satisfies next refresh`.** Documented, with the reason it is not fixed: a status frame carries no identifier for which request it answers, and the obvious heuristic (skip one frame after a timeout) makes the commoner case worse — if the earlier reply never arrives, the next refresh discards its own and times out too.

### What each round actually caught

| round | finding | outcome |
|---|---|---|
| audit 1 | timeout can expire before the request is even sent (parked behind an exclusive op) | documented in-PR, behaviour deliberately unchanged |
| audit 1 | **unsound refutation** — skeptic tested only single-refresh usage | concurrent case serialized |
| audit 2 | — | PASS; soundly refuted the `IStreamingDevice` asymmetry |
| Qodo 2 | a throwing subscriber could block the refresh | waiter taken off the public multicast event |
| Qodo 2 | a boundary reply became a timeout | checks `IsCompletedSuccessfully` first |
| Qodo 3 | late reply satisfies the next refresh | documented — no protocol correlation exists |
| audit 3 | — | **PASS, clean** |

Three of those would have told a caller the refresh **failed while the status it asked for was already applied** — the most confusing thing this API could do. Each fix is mutation-proved; one of my mutation attempts was itself invalid (it did not compile, so it proved nothing) and only surfaced because the run printed no pass/fail line.

Six tests, all of which fail without the code they cover.

### Two things filed rather than folded in

- **[#560](https://github.com/daqifi/daqifi-core/issues/560)** — `RaiseClassifiedEvent` does not isolate subscribers from each other for **any** classified event. This PR fixed its own instance by leaving the public event; the general fix allocates per frame on a 100 Hz path and needs its own measurement.
- **[#559](https://github.com/daqifi/daqifi-core/issues/559)** — a separate intermittent test failure on unmodified `main`, found while confirming #531's flake was gone. It corrected my own reporting: the baseline is 4 deterministic failures **plus** at least one intermittent, not a flat 4.

Baseline unchanged throughout: the same 4 pre-existing `LinuxUsbPortDescriptorProviderTests` failures, passing 3538 → 3544.

Issue #559, opened 2026-08-20
test(flake): LiveSampleStreamTests.ConcurrentEnumerations_AllEndOnTheSameDrop fails intermittently on a full run


cptkoolbeenz:
Found while verifying that [#531](https://github.com/daqifi/daqifi-core/issues/531)'s flake was actually gone. It is a **different test**, and it is a genuinely intermittent one.

## Observation

`Device.Internal.LiveSampleStreamTests.ConcurrentEnumerations_AllEndOnTheSameDrop` failed in **1 of 4** consecutive full-suite runs on `main` at `4791235`, with nothing else changing between runs:

| run | failures |
|---|---|
| 1 | 4 × `LinuxUsbPortDescriptorProviderTests` + **`ConcurrentEnumerations_AllEndOnTheSameDrop`** |
| 2 | 4 × `LinuxUsbPortDescriptorProviderTests` |
| 3 | 4 × `LinuxUsbPortDescriptorProviderTests` |
| 4 | 4 × `LinuxUsbPortDescriptorProviderTests` |

It did not reproduce in five consecutive runs of its own class in isolation, which is the same shape #531 had.

## Why it is filed separately from #531

#531 is `AcquisitionStatisticsTests.Record_AfterTheFirstSamplePerChannel_AllocatesNothing`, and that one is now **fixed** — root-caused as warmup-dependent (tiered-JIT promotion landing inside the measured window), fixed in [#554](https://github.com/daqifi/daqifi-core/pull/554), and green across 9 runs since. This is a different test with a different mechanism, so folding it into #531 would muddle a closed story.

## Why it matters more than a single red run

The practical cost is that **"the baseline is 4 failures" is not quite true** — it is 4 deterministic plus at least one intermittent. Anyone doing what I have been doing this session (diff a full run against a known baseline to decide whether their change broke something) will eventually see 5 and have to work out which one is theirs. I hit exactly that and had to spend four runs establishing it was not mine.

## What I have not done

I have not root-caused it. The name suggests a concurrency assertion about several enumerators observing the same drop, which is the kind of thing that goes red under scheduling pressure from a full parallel run — but that is a guess from the name, not a diagnosis, and I would rather say so than dress it up.

Worth noting the likely shape though: like #531, it passes in isolation and fails under a full run, so the trigger is probably **load or ordering**, not the assertion being wrong. #531 turned out not to be an ordering problem at all despite looking like one, so that intuition is worth testing rather than trusting.

## Reproduction

[...]

PR #561, opened 2026-08-20
fix(discovery): a device whose port you are holding is not a device that went away

### LINKED ISSUE #532: fix(discovery): a connected device is reported lost while your own app holds its serial port open
## What happens

If your app keeps discovery running while it also has a device connected — which is the normal shape for a desktop app with a device list on screen — the device you are actively using disappears from discovery. `ContinuousDeviceFinder` raises `DeviceLost` for it, and a one-shot `SerialDeviceFinder.DiscoverAsync` returns zero devices. When you disconnect, it comes straight back.

Nothing is wrong with the device. It is connected, streaming, and healthy the whole time. The only reason discovery cannot see it is that discovery cannot open the serial port, because your own process already has it open.

For a user this reads as the device dropping off and reappearing for no reason, and any UI driven by `DeviceLost` will tear down or grey out a device that is working fine.

## Why

`SerialDeviceFinder` identifies a device by opening its port and probing it with SCPI. When the port is already open, `SerialPort.Open()` throws, and the probe returns `null` — the same answer it gives for "there is no DAQiFi device on this port".

The conflation is explicit in `src/Daqifi.Core/Device/Discovery/SerialDeviceFinder.cs:665`:

```csharp
catch (UnauthorizedAccessException)
{
    // Port is in use by another application
    return null;
}
```

So Core already knows the difference between "in use" and "not a DAQiFi device" at exactly the right place — it just throws that information away. `ContinuousDeviceFinder` then counts the `null` as a miss, and after `MissThreshold` consecutive passes raises `DeviceLost`.

Worth noting the port itself is never in doubt: it stays enumerated, and the platform descriptor providers (macOS `ioreg`, Linux sysfs, Windows WMI) keep matching it, because none of them open the port. Only the SCPI probe fails. That is what makes this fixable — the candidate is still in hand when the probe fails.

## Reproduction

Bench Nq1, fw 3.7.2, macOS, over USB serial, using the example CLI. Two processes, but a single process connecting and watching behaves identically — the port lock is per-open, not per-process.

Watcher (`--watch-serial`, 1 s interval, `MissThreshold` 2) alongside a 12 s 100 Hz stream on the same device:

```
09:59:22  [+] discovered Nq1 SN:9090539562006014104 (/dev/cu.usbmodem1101)
09:59:31  (stream connects, port opened)
09:59:33  [-] lost       Nq1 SN:9090539562006014104 (/dev/cu.usbmodem1101)
09:59:48  (stream ends, port released — exit 0, ~950 samples, clean)
09:59:49  [+] discovered Nq1 SN:9090539562006014104 (/dev/cu.usbmodem1101)
```

Control: the same watcher running alone for 20 s produces one `[+]` and zero `[-]`, so the flap is caused by the port being held, not by watcher instability.

One-shot discovery while the port is held shows the same thing, and shows the port surviving enumeration:

```
Discovered 0 DAQiFi device(s):
  (no DAQiFi devices found)

Available serial ports (not verified as DAQiFi devices):
  ...
  - /dev/cu.usbmodem1101      <-- still enumerated, just not probeable
```

The same run after the port is released finds `Nq1 ... FW:3.7.2` normally.

## Suggested direction

The information needed is already at the catch site; the question is how far to carry it.

The smallest honest fix is to let the probe distinguish "port present but locked" from "no device here", and have `ContinuousDeviceFinder` treat a locked port hosting an already-known device as a hit rather than a miss. That keeps a device the caller is actively using in the live set, and leaves genuine removals — where the port stops enumerating altogether — reporting `DeviceLost` exactly as they do now.

Open questions worth deciding before implementing:

- Should a one-shot `DiscoverAsync` also report locked ports, and if so as what? It has no prior identity to reuse, so it can only say "a DAQiFi-looking port is here but busy" — which may be worth surfacing, or may be more than the API should promise.
- Windows raises different exception types for a locked port than macOS/Linux, so the detection needs to be checked per-platform rather than keyed on `UnauthorizedAccessException` alone.

Flagging rather than fixing: the loop is at its open-PR cap, and the second question above wants a call from you on how much the discovery API should promise.

Found during a bench-validation pass on `--watch-serial`, which had never been exercised against real hardware.

cptkoolbeenz:
Fixes #532.

Keep discovery running next to a connected device — the normal shape for an app with a device list on screen — and **the device you are using disappears from it.** `DeviceLost` fires, any UI driven by it tears the device down, and when you disconnect it comes straight back.

Nothing is wrong with the device. Discovery just cannot open its serial port, because your own process already has it open.

## The conflation

`SerialDeviceFinder` identifies a device by opening its port and probing it. When the port is already open, `Open()` throws and the probe returns `null` — **the same answer it gives for "there is no DAQiFi device here"**. `ContinuousDeviceFinder` counts that `null` as a miss and, after `MissThreshold` passes, calls the device lost.

The information was already at the catch site and thrown away. It is now carried.

## The fix

`SerialDeviceFinder` records ports it found present but locked; `ContinuousDeviceFinder` treats a **tracked** device on one of those as still there rather than as a miss.

**It only rescues a device already in the live set.** Its identity came from a pass that probed it successfully while the port was free, so nothing is invented. A genuine removal takes the port out of enumeration entirely, no busy report mentions it, and `DeviceLost` fires exactly as before.

That is the risk worth naming: the way this fix could go wrong is by making devices **immortal**. Two of the five tests exist for that and pass either way by design.

## The two open questions, answered

**"Should a one-shot `DiscoverAsync` also report locked ports?"** — Not decided here, deliberately. The report is carried on a separate **internal** seam (`IBusyPortReporter`) rather than by changing `IDeviceFinder.DiscoverAsync`. A one-shot discovery has no prior identity to reuse, so all it could say is *"something DAQiFi-shaped is here but busy"* — worth deciding on its own merits, and not a decision this fix has to make. Continuous discovery **has** the prior identity, which is what lets it resolve the ambiguity without guessing.

**"Windows raises different exception types for a locked port"** — **It does not.** Verified on this station rather than assumed: opening a COM port twice with `System.IO.Ports.SerialPort` under Windows threw

```
System.UnauthorizedAccessException: Access to the port 'COM12' is denied.
```

the same type as macOS and Linux. So the existing `catch (UnauthorizedAccessException)` is the right discriminator everywhere, and this needs no platform matrix. That is recorded in the code comment so the next person does not have to re-derive it.

## Test plan

| test | fails without the fix? |
|---|---|
| `DeviceOnABusyPort_IsNotLost` | **yes** |
| `BusyPortMatching_IsCaseInsensitive` | **yes** |
| `BusyPortRescue_DoesNotStopAGenuineLossLater` | **yes** |
| `DeviceWhosePortIsGone_IsStillLost` | no — control against immortality |
| `WithNoBusyReport_BehavesExactlyAsBefore` | no — control for WiFi / third-party finders |

Mutation-proved: disabling the rescue fails exactly the three that assert it, and neither control.

**Baseline-checked:** `main` fails 4 tests on this station before any change (`LinuxUsbPortDescriptorProviderTests`, a WSL/Windows runtime artefact). After: the **same 4**, passing 3546 → 3551.

## What is not covered

No test drives `SerialDeviceFinder`'s own busy-port recording — that needs a real locked port, and the unit suite has no serial hardware. The bench evidence in the issue is the coverage for that half; the `Reconcile` tests cover what the recording is *for*.

### DIFF
--- a/src/Daqifi.Core.Tests/Device/Discovery/AllTransportsDeviceFinderTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Discovery/AllTransportsDeviceFinderTests.cs
@@ -282,7 +282,7 @@ namespace Daqifi.Core.Tests.Device.Discovery
             var reporter = composite as IBusyPortReporter;
 
             Assert.NotNull(reporter);
-            var ports = reporter!.BusyPortsFromLastPass;
+            var ports = reporter!.TakeBusyPortsFromLastPass();
             Assert.Equal(new[] { "COM3" }, ports.Select(p => p.PortName));
             Assert.Equal("usb:1-1.2", ports.Single().LocationKey);
         }
@@ -299,7 +299,7 @@ namespace Daqifi.Core.Tests.Device.Discovery
                 new BusyReportingFinder("COM7", "usb:2-1"),
             });
 
-            var ports = ((IBusyPortReporter)composite).BusyPortsFromLastPass;
+            var ports = ((IBusyPortReporter)composite).TakeBusyPortsFromLastPass();
 
             Assert.Equal(new[] { "COM3", "COM7" }, ports.Select(p => p.PortName).OrderBy(n => n));
         }
@@ -315,7 +315,7 @@ namespace Daqifi.Core.Tests.Device.Discovery
             IDeviceFinder outer = new AllTransportsDeviceFinder(
                 new IDeviceFinder[] { new ListDeviceFinder(Enumerable.Empty<IDeviceInfo>()), inner });
 
-            var ports = ((IBusyPortReporter)outer).BusyPortsFromLastPass;
+            var ports = ((IBusyPortReporter)outer).TakeBusyPortsFromLastPass();
 
             Assert.Equal(new[] { "COM9" }, ports.Select(p => p.PortName));
         }
@@ -339,7 +339,9 @@ namespace Daqifi.Core.Tests.Device.Discovery
             public Task<IEnumerable<IDeviceInfo>> DiscoverAsync(TimeSpan timeout)
                 => Task.FromResult(Enumerable.Empty<IDeviceInfo>());
 
-            IReadOnlyCollection<BusyPort> IBusyPortReporter.BusyPortsFromLastPass => new[] { _busy };
+            // Re-armed by nothing: this stub always reports, which is what makes it a stub. The
+            // single-use semantics of the real ledger are pinned in BusyPortLedgerTests.
+            IReadOnlyCollection<BusyPort> IBusyPortReporter.TakeBusyPortsFromLastPass() => new[] { _busy };
         }
 
         #endregion
--- a/src/Daqifi.Core.Tests/Device/Discovery/AllTransportsDeviceFinderTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Discovery/AllTransportsDeviceFinderTests.cs
@@ -260,6 +260,90 @@ namespace Daqifi.Core.Tests.Device.Discovery
             Name = name ?? serial,
         };
 
+        #region Busy-port forwarding (#532)
+
+        [Fact]
+        public void BusyPorts_AreForwardedFromTheConstituentThatReportsThem()
+        {
+            // ContinuousDeviceFinder asks the finder it DIRECTLY wraps for busy ports:
+            //   (_finder as IBusyPortReporter)?.BusyPortsFromLastPass
+            // and the documented way to build continuous all-transport discovery is to wrap this
+            // composite. So if the composite does not carry the signal, the cast yields null, the
+            // rescue never runs, and a device whose port the app itself holds is reported lost --
+            // #532, unfixed, on the one path most callers actually use.
+            //
+            // The `as` (rather than a direct cast) is deliberate: it is what the consumer does,
+            // and it compiles whether or not the composite implements the interface, so this test
+            // fails by ASSERTION on the unfixed code rather than by not building.
+            var reporting = new BusyReportingFinder("COM3", "usb:1-1.2");
+            IDeviceFinder composite = new AllTransportsDeviceFinder(
+                new IDeviceFinder[] { new ListDeviceFinder(Enumerable.Empty<IDeviceInfo>()), reporting });
+
+            var reporter = composite as IBusyPortReporter;
+
+            Assert.NotNull(reporter);
+            var ports = reporter!.BusyPortsFromLastPass;
+            Assert.Equal(new[] { "COM3" }, ports.Select(p => p.PortName));
+            Assert.Equal("usb:1-1.2", ports.Single().LocationKey);
+        }
+
+        [Fact]
+        public void BusyPorts_AggregateAcrossConstituentsAndIgnoreNonReporters()
+        {
+            // A composite mixes transports; only some report busy ports at all. Those that do not
+            // must be skipped rather than suppressing the ones that do.
+            IDeviceFinder composite = new AllTransportsDeviceFinder(new IDeviceFinder[]
+            {
+                new BusyReportingFinder("COM3", null),
+                new ListDeviceFinder(Enumerable.Empty<IDeviceInfo>()),                 // reports nothing -- not a reporter
+                new BusyReportingFinder("COM7", "usb:2-1"),
+            });
+
+            var ports = ((IBusyPortReporter)composite).BusyPortsFromLastPass;
+
+            Assert.Equal(new[] { "COM3", "COM7" }, ports.Select(p => p.PortName).OrderBy(n => n));
+        }
+
+        [Fact]
+        public void BusyPorts_AggregateThroughANestedComposite()
+        {
+            // Nesting needs no special handling precisely BECAUSE the composite implements the
+            // interface -- the same filter picks up an inner composite and recurses. Pinned so a
+            // later "simplification" to a concrete SerialDeviceFinder check would fail here.
+            IDeviceFinder inner = new AllTransportsDeviceFinder(
+                new IDeviceFinder[] { new BusyReportingFinder("COM9", null) });
+            IDeviceFinder outer = new AllTransportsDeviceFinder(
+                new IDeviceFinder[] { new ListDeviceFinder(Enumerable.Empty<IDeviceInfo>()), inner });
+
+            var ports = ((IBusyPortReporter)outer).BusyPortsFromLastPass;
+
+            Assert.Equal(new[] { "COM9" }, ports.Select(p => p.PortName));
+        }
+
+        /// <summary>Finds nothing, but reports one port as present-and-already-open.</summary>
+        private sealed class BusyReportingFinder : IDeviceFinder, IBusyPortReporter
+        {
+            private readonly BusyPort _busy;
+
+            public BusyReportingFinder(string portName, string? locationKey)
+                => _busy = new BusyPort(portName, locationKey);
+
+#pragma warning disable CS0067 // Interface events, unused by this stub.
+            public event EventHandler<DeviceDiscoveredEventArgs>? DeviceDiscovered;
+            public event EventHandler? DiscoveryCompleted;
+#pragma warning restore CS0067
+
+            public Task<IEnumerable<IDeviceInfo>> DiscoverAsync(CancellationToken cancellationToken = default)
+                => Task.FromResult(Enumerable.Empty<IDeviceInfo>());
+
+            public Task<IEnumerable<IDeviceInfo>> DiscoverAsync(TimeSpan timeout)
+                => Task.FromResult(Enumerable.Empty<IDeviceInfo>());
+
+            IReadOnlyCollection<BusyPort> IBusyPortReporter.BusyPortsFromLastPass => new[] { _busy };
+        }
+
+        #endregion
+
         private sealed class ListDeviceFinder : IDeviceFinder, IDisposable
         {
             private readonly IReadOnlyList<IDeviceInfo> _devices;
--- a/src/Daqifi.Core.Tests/Device/Discovery/BusyPortLedgerTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Discovery/BusyPortLedgerTests.cs
@@ -28,7 +28,7 @@ public class BusyPortLedgerTests
         ledger.Record(stalePass, Port("COM3"));      // late probe, lands after the clear
         ledger.Record(currentPass, Port("COM3"));    // this pass's own observation
 
-        Assert.Equal(new[] { "COM3" }, ledger.PortsFromLastPass().Select(p => p.PortName));
+        Assert.Equal(new[] { "COM3" }, ledger.TakePortsFromLastPass().Select(p => p.PortName));
     }
 
     [Fact]
@@ -43,7 +43,7 @@ public class BusyPortLedgerTests
         ledger.Record(currentPass, Port("COM3", "usb:1-1.2"));
         ledger.Record(stalePass, Port("COM3", "usb:9-9.9"));
 
-        var ports = ledger.PortsFromLastPass();
+        var ports = ledger.TakePortsFromLastPass();
         Assert.Equal("usb:1-1.2", Assert.Single(ports).LocationKey);
     }
 
@@ -58,7 +58,7 @@ public class BusyPortLedgerTests
 
         // A port freed since the last pass must not stay "busy" -- otherwise a device really
         // unplugged while its name was occupied would be rescued forever.
-        Assert.Empty(ledger.PortsFromLastPass());
+        Assert.Empty(ledger.TakePortsFromLastPass());
     }
 
     [Fact]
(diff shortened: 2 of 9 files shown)


cptkoolbeenz:
## Audit round 1: `ready_to_merge` with one `defer_ticket`. I did not defer it.

The finding is **exactly the failure I told you the tests were guarding against** — a device made immortal — reached by a path my tests did not model. Shipping with that known-open would have made the "controls against immortality" claim in this PR body false.

### The gap

An OS port name is a **lease, not an identity**. Unplug SN-1 and the next device along can be handed `COM3`. If something then holds that new port open, matching on the **name alone** keeps SN-1 reported as present for as long as the name stays occupied, and `DeviceLost` never fires for a device that genuinely went away.

My own rationale comment claimed the invariant *"a genuine removal takes the port out of enumeration entirely"* — port-name reuse is precisely the case that breaks it, and I had not thought of it.

### Why I did not take the arbiter's first suggestion

It offered **capping consecutive rescues** or **correlating on a stronger identity**.

**A cap does not work here.** The legitimate case — your own app holding the device it is actively using — is **unbounded in time**. Any cap eventually evicts a device sitting right there, which is #532. Capping would trade the reported bug for a slower version of itself, so it is not a smaller fix, it is a wrong one.

### What it took instead

Correlating works, and the machinery already existed: `IUsbLocationProvider.GetLocationKey` resolves a **USB physical location from the OS, not from the port** — so it is available on exactly the path where the port cannot be opened.

The busy record now carries `(PortName, LocationKey)`, and a match requires the location to agree **whenever both sides know it**. A reused name behind a different physical port no longer vouches for the old device.

Where either side cannot resolve a location — an older platform provider, a port the OS will not place — it falls back to the name, which is what the previous revision did. **Never worse than before locations were carried.**

### Tests

| test | fails without the location check? |
|---|---|
| `PortNameReusedByADifferentDevice_StillReportsTheOriginalLost` | **yes** — mutation-proved by deleting the comparison |
| `SameLocationStillBusy_IsStillRescued` | no — the complement, so the check cannot degrade into "never rescue" |
| `WhenALocationIsUnknown_FallsBackToThePortName` | no — pins the fallback |

The five earlier tests keep their intent, migrated to the `BusyPort` shape.

Baseline unchanged: same 4 pre-existing failures, passing 3551 → 3554.

Round 2 running. I have asked it specifically whether a device can **still** be made immortal at this head — both sides reporting a null location, or a location key that stays stable when a different device is plugged into the same hub position.

cptkoolbeenz:
## Qodo pass 2: two bugs, both real. The first one is me over-generalising from my own measurement.

### "Access is denied" does not mean the port is still there

I claimed in a code comment that `UnauthorizedAccessException` is *"the right discriminator on every platform we support"*, citing my Windows test.

That test showed a **locked** port raises this exception. **It never showed the converse** — and this repository already documents the converse being false. `SerialPortConnectException`'s own remarks:

> Measured on macOS with System.IO.Ports 10.0.10, a missing port and a port held by another process produce **byte-identical exceptions** — same type, same message

**macOS is the platform #532 was reproduced on.** So the generalisation was wrong exactly where it mattered most: a port unplugged between enumeration and `Open()` would be recorded busy, suppressing the genuine miss it should have counted.

This is a good correction to take on the chin. I did run a real experiment, and I reported it accurately — then drew a conclusion from it that the evidence did not support, in a repo that had already measured the opposite and written it down.

**Fixed the way this library already solved it.** `SerialPortConnectException.Classify` takes a `portPresent` fact precisely because *"the platform exception cannot be classified after the fact"*. The busy path now asks the OS separately instead of reading the exception, and the comment no longer claims more than the measurement supports.

### An abandoned probe could contaminate a later pass

Also real, and I had spotted this one independently while waiting on the audit — a probe abandoned on the hard per-port timeout keeps running and can reach its catch *after* its own pass returned and after the next pass cleared the set.

Clearing could never fix it on its own: the late write lands **after** the clear. Every write now carries the pass it was started in — captured at probe **entry**, not at write time, which is the part that matters — and reads ignore anything else.

### What is not covered, stated plainly

Neither fix is unit-tested. Both live in the part of `SerialDeviceFinder` this PR already flagged as uncovered: reaching them needs a real `UnauthorizedAccessException` from `SerialPort.Open()`, which needs a genuinely locked port and the unit suite has no serial hardware. The `Reconcile`-side tests cover what the busy report is *for*; nothing covers how it is produced.

I would rather say that than add a test that exercises a fake and reads like coverage.

Baseline unchanged: same 4 pre-existing failures, passing 3554.

Audit round 2 is running against the previous head; I will re-run it against this one.

cptkoolbeenz:
## Pass 3 disposition — both bugs fixed, and they were the same bug

Both findings reduce to one fact I had flagged to the audit myself: **a USB location key identifies a physical *position*, not a device.**

### 2. Unknown location enables immortality — **fixed**, and this was the real one

This is the failure the whole mechanism must not cause: discovery going permanently blind to a removal. When neither side can offer a location, matching falls back to the port *name*, and an OS port name is a lease rather than an identity — so the rescue had no evidence about the device at all, and no bound either.

It is now bounded. A **location-confirmed** rescue still holds indefinitely, because the case it exists for — your own app holding the device it is using — has no time limit. A **name-only** rescue gets `MaxWeakBusyRescues` passes and then falls back to ordinary miss counting: exactly the behaviour before any of this existed. A blanket cap would have been the wrong fix, because it would have re-broken #532 for the case the fix was written for.

**One bug found in my own fix, by the test for it.** The first version reset the budget when it ran out, which made it *cycle* — twenty rescues, one miss, reset, twenty more. The device was still never lost, so the bound bought nothing. The budget now resets only on a real sighting.

### 1. Location aliases a replacement device — **acknowledged, inherent, blast radius capped**

Correct, and not fixable at this layer. The only true identity is the serial number, and reading it requires opening the port — which is precisely what a busy port refuses. Any port-level identity check has this residual.

What changed is that the code no longer *implies* the key is an identity: the comment says it is a position, and the weak half is bounded, so the worst case is a stale entry for a bounded number of passes rather than forever. The consequence is also mild in the direction that matters — a replacement device in the same slot is reported present, and it *is* present.

### Tests

Three added, each mutation-checked:

| test | pins |
|---|---|
| `Reconcile_NameOnlyRescue_IsBounded` | past the bound, `DeviceLost` fires — this is the one that caught the cycling bug |
| `Reconcile_LocationConfirmedRescue_IsNotBounded` | 200 passes, still held; stops a blanket cap from creeping back in |
| `Reconcile_ARealSighting_ResetsTheWeakRescueBudget` | the budget is about *consecutive* unverifiable passes |

Suite: 3557 passed, 4 failed — the same four pre-existing `LinuxUsbPortDescriptorProviderTests` failures that `main` has.

cptkoolbeenz:
## Adversarial audit round 3 — one finding that mattered a great deal

**Finding 0 — `busyPorts` is null on the default construction path — CONFIRMED, fixed in `2c58e8f`.**

This is the finding worth having run the audit for. It says the PR did not fix #532 on the path almost everyone uses, and it is correct.

`ContinuousDeviceFinder` asks the finder it *directly* wraps:

```csharp
(_finder as IBusyPortReporter)?.BusyPortsFromLastPass
```

Only `SerialDeviceFinder` implemented that interface. But the documented way to build continuous all-transport discovery — stated in `AllTransportsDeviceFinder`'s own `<remarks>` — is to wrap `AllTransportsDeviceFinder.CreateDefault()`. The composite did not carry the signal, so the cast yielded `null`, the entire rescue block was skipped, and a device whose port the application itself was holding was **still reported lost**. The feature worked only when wrapping a bare `SerialDeviceFinder`, which is not how the class is meant to be composed.

Every test I had written passed, because they all called `Reconcile` directly and *handed it* the busy ports. The defect was in how that argument is obtained, which no test reached. That is the useful lesson here: a seam that is always supplied by the test is a seam that is never tested.

`AllTransportsDeviceFinder` now implements `IBusyPortReporter` and aggregates from whichever constituents report. Nesting needs no special handling, because the composite itself is now a reporter and the same filter recurses.

**Finding 1 — `TryAdd` is first-writer-wins across a pass boundary — fixed rather than deferred.**

The audit deferred this as low, on the reasoning that the interleaving is narrow. I fixed it instead, for two reasons: Qodo raised the same defect independently at importance 10 in the same pass, and the fix is small. It is also not as improbable as "narrow timing window" suggests — the pass tag exists *precisely because* abandoned probes finish late, so by the design's own premise the late write is expected, not exotic.

The bookkeeping moved into a `BusyPortLedger` type, which is what made it testable without a test-only hook. `Record` now uses `AddOrUpdate` with newest-pass-wins, correct in both directions: a stale write can neither block the current pass's entry (what `TryAdd` did) nor clobber it (what a plain indexer assignment would do). The pass comparison uses the wrap-safe difference form rather than `>`.

### Mutation proof

Both fixes, each mutation compiling so the result means something:

| mutation | result |
|---|---|
| composite stops implementing `IBusyPortReporter` | **3 failed** — all three forwarding tests |
| `Record` reverts to `TryAdd` | **1 failed** — exactly the race test, and none of the other nine |

The second is the one I care about: a test that fails only for its own mutation, rather than for any change at all.

Suite: 3570 passed, 4 failed — the same four pre-existing `LinuxUsbPortDescriptorProviderTests` failures that `main` has.

cptkoolbeenz:
## Pass 5 — three findings, all fixed in `2e99630`

All three are consequences of the composite forwarding I added last commit, which is a fair trade: that change fixed the feature on the path everyone actually uses, and these are the seams it opened.

**Bug — "Stale busy ports cross passes". Correct, and the same shape as the unbounded-rescue bug from pass 3:** evidence that is not evidence, held indefinitely. A timed discovery can return before a constituent ever acquired its semaphore and began a pass, so the newest data in that constituent is from an *older* moment — and forwarding it again lets repeated contention reuse one stale location-confirmed observation forever. Since that path is deliberately unbounded, a device that really was unplugged would never be reported lost.

The snapshot is now **single-use**: `BeginPass` arms it, taking it disarms it. The question it answers is not *"what was busy recently"* but *"what did the pass that just ran observe"* — and if no pass ran, the honest answer is nothing, not last time's. `IBusyPortReporter` becomes `TakeBusyPortsFromLastPass()` rather than a property, so the consumption is visible at the call site instead of hidden behind a property read.

**Suggestion (8) — "Resolve duplicate busy reports safely". Applied.** Directly caused by the aggregation: one port name can now be reported by more than one constituent, and returning on the first match let a report carrying no location downgrade the device to the bounded path even though a later report confirmed the location outright. It now considers every matching report, and treats contradictory evidence — a name-only report alongside one placing the port somewhere else — as rescuing nothing.

**Suggestion (8) — "Avoid rescuing unverifiable missing ports". Applied.** `IsPortStillPresent` failed *open*, which re-opens the exact hole the method was added to close, and feeds it into the unbounded path. Failing closed is the right asymmetry: a wrong "gone" costs one miss and `MissThreshold` absorbs it; a wrong "present" can keep a phantom alive forever. It is also the pre-PR behaviour.

### Tests — 5 added, and one honest gap

| fix | proof |
|---|---|
| single-use snapshot | mutation: making the take not disarm fails **exactly** the single-use test |
| all-reports classification | mutation: restoring the first-match return fails **2** classification tests |
| fail-closed presence | **no direct test** — see below |

The fail-closed change is a one-line fail-safe I have not pinned with a test. Reaching that catch needs a port-name provider that throws only on the *second* call within a pass, which couples the test to internal call ordering and would break on any reordering that is otherwise harmless. I would rather say that plainly than ship a brittle test and call it covered. It is argued from direction of failure, which is the same reasoning the reviewer used to raise it.

### Suite

3574 passed. The 4 pre-existing `LinuxUsbPortDescriptorProviderTests` failures are unchanged from `main`.

One further failure appeared in the full run — `AcquisitionStatisticsTests.AttachedToDevice_CostsTheDecodePathNoMoreThanAnEmptySubscriber`. It passes **3/3 in isolation** and this PR touches only `Device/Discovery`, so it is not from these changes: it is a third instance of the allocation-measurement flake already tracked as #531/#559, in the same file and with the same in-a-full-run-only signature. Noted on #559 with the diagnosis rather than folded into this PR's scope.

cptkoolbeenz:
## Pass 6 + audit round 4 — three findings, all fixed in `c1f112c`

**Audit, CONFIRMED — a rescue never cleared `MissCount`.** The best finding of this round, and it undermines the rescue's own purpose. `MissThreshold` counts **consecutive** misses — `Reconcile`'s own doc comment says so, and the default of 2 exists to tolerate one dropped response. The rescue path skipped the increment but left any earlier count in place, so `miss → rescue → miss` fired `DeviceLost` after **a single real miss following a pass in which the device was affirmatively confirmed present**. After any rescue the effective tolerance was zero.

Both branches now clear it, because in both the pass was not a miss. The weak branch still terminates — `WeakRescueCount` is the bound, and unlike `MissCount` it deliberately survives across rescues, so the budget still runs out.

**Qodo — a confirming report short-circuited the scan.** Right, and it was an inconsistency in my own previous commit rather than a new case: contradiction was already fail-safe for the name-only path, and the early return on the first location match was what skipped the check. A mismatch now poisons the whole classification, confirming reports included. It is the only *negative* evidence available here, and the asymmetry decides it — the confirmed path is unbounded, so accepting a contradiction keeps an absent device alive forever, while declining costs a miss the threshold absorbs.

**Qodo — taking the snapshot was not atomic with beginning a pass.** Correct: disarming and reading the entries were two steps, so a queued discovery could start in between and hand the taker the next pass's half-filled data while the next taker got nothing. Both now hold one lock. `Record` stays outside it — it is a `ConcurrentDictionary` write and must not serialise against the concurrent probes that call it.

### Tests

| fix | proof |
|---|---|
| `MissCount` cleared on rescue | mutation: removing the reset fails **exactly** that test |
| mismatch poisons a confirmation | mutation: restoring the early return fails **exactly** that test |
| atomic take | **no test** — see below |

The `MissCount` test also pins that the tolerance is *restored* rather than merely deferred: after the rescue it still takes a full threshold of consecutive misses.

The atomicity fix is a race, fixed by construction. A deterministic test would have to interleave the two operations at an instruction boundary, and a timing-dependent one would be worse than no test — it would pass on a broken build most runs and turn a real regression into a flake. Disclosed rather than glossed, same as the fail-closed change last round.

### Suite

3577 passed; the 4 pre-existing `LinuxUsbPortDescriptorProviderTests` failures unchanged from `main`. The allocation flake from last round did not recur this run, consistent with it being the intermittent tracked in #559.

cptkoolbeenz:
## Pass 7 — blocking tier clear; non-blocking triaged

Six of the seven bugs raised across this PR now read **✓ Resolved**. The one remaining — *"Location aliases replacement devices"* — is the finding dispositioned in round 3 as inherent, and `/agentic_review` keeps a declined finding listed on every subsequent pass rather than un-listing it. **The remaining count equals the declined set**, so the blocking tier is empty.

Restating the decline once so it is not re-derived: a USB location key identifies a physical **position**, not a device. Any port-level identity check has this residual, because the only true identity is the serial number and reading it requires *opening* the port — which is exactly what a busy port refuses. It is mitigated rather than fixed: the weak half is bounded, a mismatch now poisons the confirmed path too, and the comment states the key is a position rather than implying identity.

### Non-blocking triage — "Preserve case-sensitive POSIX port identities" (importance 3)

**Half applied, half declined**, and the half that was right is worth saying out loud.

**The observation is correct and my comment was wrong.** `BusyPortLedger` said *"an OS port name is not case-significant"*. POSIX device paths **are** case-sensitive, so that was an overstatement. Corrected in `38f0c88` to say what is actually true, and why the comparison is still safe: both the recorded name and the looked-up one come from the **same enumeration**, so they never differ by case alone. `ClassifyBusyRescue` already carried that more careful wording — the ledger's copy was the sloppy one.

**The code change is declined.** Platform-branching the comparison guards a case that cannot arise: for two entries to collide wrongly, one enumeration would have to return the same device node under two spellings. It would also introduce a divergence between recording and lookup, which is a real failure mode in exchange for a hypothetical one. If a POSIX platform is ever found that returns inconsistent casing, the fix belongs at the enumeration boundary — normalising once — not in two comparison sites that must then be kept in agreement forever.

### Merge readiness

- **Adversarial audit round 5: `gate: PASS`, clean, zero findings** against `c1f112c`. I checked the run journal rather than trusting the summary — the agent that reported nothing returned `{"findings": []}`, a genuine nothing-found rather than a dead leg.
- The only delta since that audited head is `38f0c88`, verified **comment-only** (`git diff --unified=0` shows no non-comment line on either side). Re-running a full audit for a comment would be waste; saying so is better than implying the audit covered a diff it did not see.
- Suite: **3577 passed**, and the same 4 pre-existing `LinuxUsbPortDescriptorProviderTests` failures that `main` has.

Per the anti-treadmill rule I am not re-triggering `/improve` for a decline-plus-comment pass.

Issue #644, opened 2026-08-24
chore(build): the published Daqifi.Mcp package has no license, project URL, repository URL, tags, or SourceLink


tylerkron:
Noticed while doing #638 (centralizing shared build properties). Filed separately because fixing it changes what ships in a published package, which #638 deliberately did not do.

## What

`Daqifi.Core.csproj` sets a full set of package metadata: `PackageLicenseExpression`, `PackageProjectUrl`, `PackageTags`, `RepositoryUrl`, `RepositoryType`, plus symbol/SourceLink settings (`IncludeSymbols`, `SymbolPackageFormat`, `PublishRepositoryUrl`, `ContinuousIntegrationBuild`, and a `Microsoft.SourceLink.GitHub` reference).

`Daqifi.Mcp.csproj` sets only `PackageId`, `Authors`, `Description`, and `PackageReadmeFile`. Evaluated with `dotnet msbuild src/Daqifi.Mcp/Daqifi.Mcp.csproj -getProperty:PackageLicenseExpression -getProperty:PackageProjectUrl -getProperty:RepositoryUrl -getProperty:PackageTags`, all four come back empty.

`Daqifi.Mcp` is a published package (`dotnet tool install -g Daqifi.Mcp`), packed by `release.yml` from the same tag as `Daqifi.Core`. So on nuget.org it shows no license, no project link, no source repository, and no tags, and it has no symbol package or SourceLink for anyone debugging into it.

## Why it was not folded into #638

The genuinely shared property here is `Authors` (`DAQiFi` in both). The rest is not shared today, and copying `Daqifi.Core`'s values into a shared props file would silently rewrite the `Daqifi.Mcp` package's metadata as a side effect of a "no behavior change" cleanup. That is a real decision about a shipped artifact and belongs in its own PR.

## Suggested fix

Give `Daqifi.Mcp` the same license/URL/tag/SourceLink metadata as `Daqifi.Core`, most likely by moving the values that are genuinely repo-wide (`Authors`, `PackageLicenseExpression`, `PackageProjectUrl`, `RepositoryUrl`, `RepositoryType`, and the SourceLink/symbol settings) into the root `Directory.Build.props` added by #638, guarded by `'$(IsPackable)' != 'false'` if the test projects should stay untouched. Verify by unpacking both `.nupkg` files and diffing the `.nuspec` before and after.

tylerkron:
[...]

Issue #643, opened 2026-08-24
chore(build): decide whether Daqifi.Mcp should multi-target net9.0;net10.0 like Daqifi.Core


tylerkron:
Split out of #638, which centralized the shared build properties into `Directory.Build.props` but deliberately left `TargetFramework(s)` alone.

## What

`Daqifi.Core` and `Daqifi.Core.Tests` target `net9.0;net10.0`. `Daqifi.Mcp` and `Daqifi.Mcp.Tests` target `net9.0` only.

`Daqifi.Mcp` is packed with `PackAsTool` and installed with `dotnet tool install -g Daqifi.Mcp`, so the tool package currently carries a `net9.0` asset and nothing else. A user whose machine only has the .NET 10 runtime installed has to also install .NET 9 to run `daqifi-mcp`, even though the library it wraps already supports .NET 10. `release.yml` packs both packages from the same version tag, so the asymmetry ships on every release.

## Why it was not done in #638

#638 was a no-behavior-change build-config cleanup: same compiled output, same package contents. Adding a `net10.0` asset to a published tool package changes what ships, which deserves its own PR and its own verification (build + test the MCP server on net10.0, confirm `ModelContextProtocol` and `Microsoft.Extensions.Hosting` behave the same there, and confirm the packed tool still resolves).

## What to decide

Either:

- **Multi-target** `Daqifi.Mcp` (and `Daqifi.Mcp.Tests`) to `net9.0;net10.0` to match `Daqifi.Core`. `release.yml`'s `dotnet pack` needs no change; the tool package would carry both assets. Or
- **Keep it single-TFM** and record why in the `.csproj` next to the `TargetFramework` line, so the next person does not read it as an oversight.

Whichever way it goes, the decision should be written down rather than left implicit.

Issue #484, opened 2026-08-12
chore: add .editorconfig + style enforcement; rename the two public SCREAMING_SNAKE constants


tylerkron:
**Type**: Chore
**Priority**: Low

## What
Nothing enforces code style in a repo that otherwise sets `TreatWarningsAsErrors`:

- No `.editorconfig`, `Directory.Build.props`, or ruleset exists anywhere.
- Namespace style is split: 194 files file-scoped vs 58 block-scoped in `src/Daqifi.Core` (new files copy whichever neighbor they were cloned from).
- Two SCREAMING_SNAKE constants are **public API on the published NuGet package**: `IntelHexParser.cs:14` `DEFAULT_BEGIN_PROTECTED_ADDRESS` and `:19` `DEFAULT_END_PROTECTED_ADDRESS` (five more files use the style internally).
- Bonus doc bug to fold in: `docs/DEVICE_INTERFACES.md:945` tells readers to call `ScpiMessageProducer.TurnDeviceOff` — that member has never existed (only `TurnDeviceOn`); the snippet does not compile.

## Why
Style drift compounds with every file, and the public constants get more expensive to rename with each release. An `.editorconfig` also gives later work an enforcement point (e.g. CA2007 `ConfigureAwait` on the connect path — see the companion transport-deadlock issue).

## How
1. Commit `.editorconfig` with `csharp_style_namespace_declarations = file_scoped:warning` + standard .NET naming rules; wire `EnforceCodeStyleInBuild`.
2. Rename the two public constants to PascalCase with `[Obsolete]` forwarders for the shipped names.
3. Convert the 58 block-scoped namespaces in one mechanical, review-separate PR.
4. Fix the `TurnDeviceOff` doc line.

## Success criteria
- [ ] `.editorconfig` enforced in build; CI fails on new style violations.
- [ ] PascalCase replacements shipped with `[Obsolete]` forwarders; 0 block-scoped namespaces remain.
- [ ] The `DEVICE_INTERFACES.md` snippet compiles.

PR #647, opened 2026-08-24
chore(build): centralize the shared build properties in Directory.Build.props

### LINKED ISSUE #638: chore(build): no Directory.Build.props — four csproj files drift, and Daqifi.Mcp alone has no TreatWarningsAsErrors
**Type**: Build / Consistency
**Priority**: Medium

## What
There is no `Directory.Build.props`. Four project files each redeclare the same properties by hand, and they have already drifted:

| Property | `Daqifi.Core` | `Daqifi.Core.Tests` | `Daqifi.Mcp` | `Daqifi.Mcp.Tests` |
|---|---|---|---|---|
| `TargetFramework(s)` | `net9.0;net10.0` | `net9.0;net10.0` | **`net9.0`** | **`net9.0`** |
| `ImplicitUsings` | enable | enable | enable | enable |
| `Nullable` | enable | enable | enable | enable |
| `TreatWarningsAsErrors` | true | true | **absent** | true |

Two concrete drifts fall out of that table:

1. **`Daqifi.Mcp` is the only project in the repo without `TreatWarningsAsErrors`** (`src/Daqifi.Mcp/Daqifi.Mcp.csproj`). Its own test project has it; the shipping tool does not. A new warning in `DaqifiAgent.cs` (1,701 lines) or `Tools/DaqifiTools.cs` builds green.

2. **`Daqifi.Mcp` targets `net9.0` only** while `Daqifi.Core` targets `net9.0;net10.0`. It is packed with `PackAsTool` and installed via `dotnet tool install -g Daqifi.Mcp`, so a user on a .NET 10-only machine cannot run it without also installing the .NET 9 runtime — even though the library it wraps already supports .NET 10. `release.yml` packs both packages from the same version tag, so this asymmetry ships on every release.

The `.editorconfig` has the same scoping problem. `src/Daqifi.Core/.editorconfig` enables `CA2007` — and its header says so explicitly:

```
# Analyzer configuration for the Daqifi.Core library only.
```

That is a deliberate and correct decision for CA2007 (`CONTRIBUTING.md` documents test projects as exempt). But because it is the *only* `.editorconfig` in the repo, there is nowhere to put settings that genuinely should apply everywhere — formatting, naming, `dotnet_style_*` — and `Daqifi.Mcp`, which is shipped code, gets no analyzer configuration at all.

## Why
Each drift is individually small; the pattern is that nothing prevents the next one. Every new project starts by copy-pasting the previous one's `PropertyGroup`, and any repo-wide decision (bumping a TFM, enabling an analyzer, adding SourceLink, turning on deterministic builds) has to be applied four times and is silently incomplete if it is applied three.

The TFM drift is the one with user-visible consequence today, and it is the kind of thing that gets noticed as a support question rather than in review.

This is adjacent to but distinct from **#484** (add `.editorconfig` + style enforcement; rename the two public SCREAMING_SNAKE constants). #484 is about *which style rules* to adopt; this is about *where build properties live* so a rule can be adopted once. Doing this first gives #484 somewhere to land.

## How
1. Add `Directory.Build.props` at the repo root with the shared properties: `ImplicitUsings`, `Nullable`, `TreatWarningsAsErrors`, and the common package/repository metadata. Remove the now-duplicated lines from the four `.csproj` files, leaving each with only what is genuinely project-specific (`PackageId`, `OutputType`, `PackAsTool`, `IsPackable`, coverlet settings).
2. Decide the `Daqifi.Mcp` TFM question explicitly rather than by omission — either multi-target it to `net9.0;net10.0` to match Core, or record why the tool stays single-TFM. If it multi-targets, `release.yml`'s `dotnet pack` needs no change, but the tool package will carry both.
3. Move the repo-wide `.editorconfig` to the root with `CA2007` kept scoped to the library via a `[src/Daqifi.Core/**.cs]` section, so the existing exemption for test projects survives verbatim and `Daqifi.Mcp` starts getting analyzer coverage.

Do not centralize `TargetFrameworks` itself into `Directory.Build.props` until step 2 is settled — projects legitimately differ there and a shared default would hide that rather than fix it.

## Success criteria
- [ ] `Directory.Build.props` at the repo root; no property in the table above is declared in more than one place.
- [ ] `Daqifi.Mcp` builds with `TreatWarningsAsErrors` and is warning-clean.
- [ ] The `Daqifi.Mcp` TFM set is a recorded decision, not an omission.
- [ ] `.editorconfig` at the root, with the CA2007 scoping and the test-project exemption from `CONTRIBUTING.md` preserved exactly.
- [ ] `dotnet build Daqifi.Core.sln` and `dotnet test Daqifi.Core.sln` unchanged in behavior.

tylerkron:
## What was wrong

The repo had no `Directory.Build.props`, so all four project files declared the same three build settings by hand — and they had already drifted apart. `Daqifi.Mcp`, the MCP server we publish as a `dotnet tool`, was the only project in the repo without `TreatWarningsAsErrors`. Every other project, including `Daqifi.Mcp`'s own test project, treats warnings as errors; the shipping tool did not, so a new compiler warning in `DaqifiAgent.cs` or `Tools/DaqifiTools.cs` would have built green and shipped. Nothing prevented the next drift either: each new project starts life as a copy-paste of the previous one's `PropertyGroup`, and any repo-wide decision has to be remembered in four places.

## How it was fixed

`ImplicitUsings`, `Nullable` and `TreatWarningsAsErrors` now live once in a root `Directory.Build.props`, and the per-project copies are gone. That closes the `Daqifi.Mcp` gap as a side effect — it was already warning-clean in both Debug and Release, so nothing had to be fixed to turn the flag on. The CA2007 `.editorconfig` moves to the repo root as well, with its section changed from `[*.cs]` to `[src/Daqifi.Core/**.cs]` so the library keeps the rule and the test-project exemption documented in `CONTRIBUTING.md` survives verbatim; the root is simply where a repo-wide rule can now go (which is what #484 will need).

Two things a reviewer might expect and won't find, both deliberate:

- **`TargetFramework(s)` is not centralized.** `Daqifi.Core` multi-targets `net9.0;net10.0` and `Daqifi.Mcp` is `net9.0` only. That is a real asymmetry with a user-visible consequence (a .NET 10-only machine cannot run `dotnet tool install -g Daqifi.Mcp` without also installing .NET 9), but resolving it changes what the released package contains, which does not belong in a no-behavior-change cleanup. It is now filed as **#643** and referenced from a comment next to the `TargetFramework` line in both MCP projects, so it is a recorded decision rather than an omission.
- **Package metadata is not centralized.** `Daqifi.Core` sets license, project URL, repository URL, tags and SourceLink; `Daqifi.Mcp` sets none of them. Only `Authors` is genuinely common, so hoisting the rest would have silently rewritten the published `Daqifi.Mcp` package's metadata. Filed as **#644**.

## Verification

- **Before/after evaluated property sets are identical except for the one intended change.** Dumped ~55 properties per project with `dotnet msbuild <csproj> -getProperty:<Name>` for all ten (project, target-framework) pairs, on `origin/main` and on this branch with restore warm on both sides, and diffed. The only difference anywhere is `Daqifi.Mcp`'s `TreatWarningsAsErrors`: `false` → `true`.
- **CA2007 scoping was tested, not assumed.** Temporarily dropping one `.ConfigureAwait(false)` in `TcpStreamTransport.cs` fails the build with `error CA2007`, and the test projects — which contain plenty of naked `await`s — still build clean.
- Full suite green on both frameworks: net9.0 (3820 Core + 217 MCP) and net10.0 (3820 Core). Release build of the solution: 0 warnings, 0 errors.
- The example CLI at `daqifi-core-example-app` builds and runs against this branch's `Daqifi.Core`, and streamed for 5 s from a real Nyquist over `/dev/cu.usbmodem1101`.

New tests (`DirectoryBuildPropsTests`) fail the build if any project redeclares a property that `Directory.Build.props` owns, which is the exact regression that produced this issue. They were confirmed to fail when the drift is reintroduced.

closes #638

---

Not merging — for review.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,11 @@
-# Analyzer configuration for the Daqifi.Core library only.
+# Repo-wide analyzer and editor configuration.
+#
+# This file lives at the repository root so that a rule meant for the whole repo
+# has somewhere to go (issue #638). Rules that are deliberately scoped to one
+# project stay scoped, via a path-qualified section below.
+root = true
+
+# CA2007 (ConfigureAwait) applies to the Daqifi.Core library only.
 #
 # Daqifi.Core ships synchronous facades that block on their own async work
 # (DaqifiDevice.Connect(), DaqifiDeviceFactory.ConnectTcp/ConnectSerial,
@@ -10,6 +17,8 @@
 #
 # CA2007 makes that a build error rather than a code-review habit. Library code
 # has no business resuming on the caller's context, so the rule applies to every
-# await in this project, not just the connect path.
-[*.cs]
+# await in this project, not just the connect path. Test projects are exempt
+# (CONTRIBUTING.md), which is why this section is path-qualified instead of
+# applying to every [*.cs] in the repo.
+[src/Daqifi.Core/**.cs]
 dotnet_diagnostic.CA2007.severity = warning
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ All code changes go through a pull request:
 `DaqifiDeviceFactory.ConnectTcp`/`ConnectSerial`, `IStreamTransport.Connect`/`Disconnect`)
 that block on their own async work. An `await` that resumes on the caller's
 `SynchronizationContext` therefore deadlocks a WPF/WinForms app calling from the UI
-thread. CA2007 is enabled for the library project (see `src/Daqifi.Core/.editorconfig`)
+thread. CA2007 is enabled for the library project (see the `[src/Daqifi.Core/**.cs]` section of the root `.editorconfig`)
 and warnings are errors, so a naked `await` fails the build. Test projects are exempt.
 
 ## Security: how we do and don't accept code
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,30 @@
+<Project>
+
+  <!--
+    Repo-wide build properties. MSBuild imports this file into every project
+    under the repository root, so a setting here is declared once instead of
+    being copy-pasted into each .csproj (issue #638).
+
+    Only settings that every project in the repo should share belong here.
+    Anything a project legitimately decides for itself - PackageId, OutputType,
+    PackAsTool, IsPackable, coverlet output, package metadata - stays in that
+    project's .csproj.
+
+    TargetFramework(s) is deliberately NOT centralized: the projects genuinely
+    differ (Daqifi.Core and Daqifi.Core.Tests multi-target net9.0;net10.0, while
+    Daqifi.Mcp and Daqifi.Mcp.Tests are net9.0 only), and a shared default here
+    would hide that difference rather than resolve it. See issue #643 for the
+    Daqifi.Mcp target-framework decision.
+  -->
+  <PropertyGroup>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <!--
+      Warnings are errors everywhere, including Daqifi.Mcp, which was the one
+      project that never opted in. It is warning-clean in Debug and Release, so
+      turning it on here needed no code changes.
+    -->
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+
+</Project>
--- a/src/Daqifi.Core.Tests/Build/DirectoryBuildPropsTests.cs
+++ b/src/Daqifi.Core.Tests/Build/DirectoryBuildPropsTests.cs
@@ -34,13 +34,17 @@ public class DirectoryBuildPropsTests
     /// Property names declared in the root <c>Directory.Build.props</c>, read from the file itself
     /// so that adding a property there automatically extends the coverage below.
     /// </summary>
-    private static IReadOnlyList<string> CentralizedPropertyNames() =>
+    /// <remarks>
+    /// Compared case-insensitively because MSBuild property names are: <c>&lt;Nullable&gt;</c> and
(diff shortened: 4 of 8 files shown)


tylerkron:
Good catch on the case-sensitivity — conceded and fixed in 11fb0c8. MSBuild property names are case-insensitive, so `<nullable>` overrides the centralized `<Nullable>` exactly as an exact-case redeclaration would, and the ordinal comparison would have let it through. `CentralizedPropertyNames()` now returns a `HashSet<string>` built with `StringComparer.OrdinalIgnoreCase`, which the offender scan uses for its lookup too.

Verified rather than assumed: adding `<nullable>enable</nullable>` to `Daqifi.Mcp.Tests.csproj` now fails with `src/Daqifi.Mcp.Tests/Daqifi.Mcp.Tests.csproj declares <nullable>`, where before the fix it passed. Added `CentralizedPropertyNames_AreMatchedCaseInsensitively` to pin the comparer so it cannot regress to ordinal.

On the alternative approaches: agreed that a `Directory.Build.targets` check would fail earlier than a test. Not doing it here — it would add MSBuild machinery to a PR whose whole point is that the build config should be simple to read, and CI runs `dotnet test` on every PR so the guard does fire. Worth revisiting if the repo ever gains a build path that skips tests.

Full suite re-run green on both frameworks: net9.0 3821 Core + 217 MCP, net10.0 3821 Core.

tylerkron:
Qodo round 2 is clean on head `11fb0c8`: Bugs (0), Rule violations (0), the case-sensitivity finding marked Resolved, and no unresolved inline review threads. Re-checked both surfaces after a settle interval and they are still clean on the same SHA. CI green.

Ready for review — not merging.

PR #649, opened 2026-08-24
fix(scpi): reject a negative channel on the two DIO producers

### LINKED ISSUE #624: ScpiMessageProducer: SetDioPortDirection and SetDioPortState accept a negative channel while the other nine channel commands reject it
Noticed while unifying the negative-channel guard in `ScpiMessageProducer` (PR to follow). Filing separately because fixing it is a behavior change to public methods, not a refactor.

## The asymmetry

Eleven public `ScpiMessageProducer` methods take an `int channel`. Nine of them open with the same guard:

```csharp
if (channel < 0)
{
    throw new ArgumentOutOfRangeException(nameof(channel), channel, "Channel number cannot be negative.");
}
```

Two do not:

| Method | Guards negative channel? |
| --- | --- |
| `SetPwmChannelEnabled` | yes |
| `SetPwmChannelFrequency` | yes |
| `SetPwmChannelDutyCycle` | yes |
| `SetAnalogOutputVoltage` | yes |
| `GetAnalogOutputVoltage` | yes |
| `SetAdcCalibrationSlope` | yes |
| `SetAdcCalibrationOffset` | yes |
| `GetAdcCalibrationSlope` | yes |
| `GetAdcCalibrationOffset` | yes |
| **`SetDioPortDirection`** | **no** |
| **`SetDioPortState`** | **no** |

`SetDioPortDirection(-1, 1)` returns a perfectly well-formed message object whose payload is `DIO:PORt:DIRection -1,1`, and `SetDioPortState(-1, 1)` likewise produces `DIO:PORt:STATe -1,1`. Both go to the wire. Every sibling command would have thrown on the same input.

## Why it matters

The producer is public API, so a caller writing against it sees an inconsistent contract: the same illegal argument is a caught programming error on nine commands and a silently malformed SCPI frame on two. A negative channel is never valid for the firmware, so the DIO case is a bug that surfaces late (as a device-side rejection or a no-op) instead of at the call site.

`ScpiMessageProducer.SetDioPortDirection` also takes `int direction` with no validation, while the device-level `ChannelControlOperations` validates direction is Input or Output before calling it. Same shape of gap, worth deciding at the same time.

## Options

1. **Add the same guard to both DIO methods** (recommended). Makes the eleven consistent and matches what the rest of the class already promises. It is a behavior change: a caller currently passing a negative channel would start getting `ArgumentOutOfRangeException` instead of a bad message. That seems like the point.
2. Add it to the device-level wrappers only, leaving the producer permissive. Weaker -- the producer is public, so callers can reach it directly.
3. Document the two as deliberately unvalidated. Only defensible if some caller genuinely needs to emit a negative channel, which seems unlikely.

Recommend option 1, done as its own change so the behavior shift is visible in the log rather than folded into a no-behavior-change refactor.

Produced by the duplicate-unifier maintenance routine.

tylerkron:
## What was wrong

`ScpiMessageProducer` has eleven public commands that take a channel number. Nine of them reject a negative channel immediately, with an `ArgumentOutOfRangeException` naming the bad argument. The two digital-I/O commands did not: `SetDioPortDirection(-1, 1)` cheerfully handed back a perfectly well-formed message whose payload was `DIO:PORt:DIRection -1,1`, and `SetDioPortState(-1, 1)` did the same. That frame went out to the device, which has no such channel — so a plain programming mistake surfaced late as a device-side rejection or a silent no-op, instead of at the line that made it. Worse, the library was telling callers two different stories about the same illegal argument depending on which command they happened to be using.

## How it was fixed

Both DIO commands now call the same `ValidateChannel` helper the other nine already share, so all eleven throw the identical exception, with the same `ParamName` and the same message. On `SetDioPortState` the channel check runs ahead of the finite-value check added in #645, so a caller who gets both arguments wrong is told about the first one.

The thing worth pushing back on: **this rejects input that used to be accepted.** Code that passed a negative channel to either DIO command previously got a message object and now gets an exception. That is the point of the change — a negative channel was never valid for the firmware — but it is a behavior change to public API, which is why it is its own commit rather than folded into a refactor. Nothing inside this repo is affected: the only production callers are the device-level DIO wrappers, which pass a channel number resolved from the device's own channel collection and so can never be negative. Channel `0` remains valid and is pinned by a test.

## How it was verified

- A throwaway probe first confirmed both methods really did accept `-1` today, and the pre-existing nine-command uniformity test was confirmed green against unchanged code — so the new assertions are checked against a proven reference rather than a copied literal.
- The four new-behavior assertions were then run against unchanged code and failed, as they must, before the two-line production change was made.
- The existing `EveryChannelAddressedProducer_RejectsNegativeChannelIdentically` test now covers all eleven commands instead of nine, so a twelfth channel command added without the guard fails the suite.
- Clean `--no-incremental` solution build: 0 warnings, 0 errors. Full `dotnet test` green on net9.0 and net10.0 (3832 passed per TFM, plus 217 in Daqifi.Mcp.Tests).
- Bench: smoke-tested against the board over USB with the example CLI built against this branch. The CLI exposes no DIO command, so these two producers cannot be driven on the wire through it; for valid channels the emitted bytes are unchanged either way, so there was nothing hardware could add here.

closes #624

**Not merging — for review.**

### DIFF
--- a/src/Daqifi.Core.Tests/Communication/Producers/ScpiMessageProducerTests.cs
+++ b/src/Daqifi.Core.Tests/Communication/Producers/ScpiMessageProducerTests.cs
@@ -181,11 +181,53 @@ public class ScpiMessageProducerTests
     }
 
     [Fact]
+    public void SetDioPortDirection_WithNegativeChannel_Throws()
+    {
+        Assert.Throws<ArgumentOutOfRangeException>(() => ScpiMessageProducer.SetDioPortDirection(-1, 1));
+    }
+
+    // Channel 0 is a real channel on every DAQiFi board, so the guard's boundary matters:
+    // rejecting only negatives must leave the lowest valid channel untouched.
+    [Fact]
+    public void SetDioPortDirection_WithChannelZero_StillReturnsCommand()
+    {
+        var message = ScpiMessageProducer.SetDioPortDirection(0, 0);
+        Assert.Equal("DIO:PORt:DIRection 0,0", message.Data);
+        AssertMessageFormat(message);
+    }
+
+    [Fact]
     public void SetDioPortState_ReturnsCorrectCommand()
     {
         var message = ScpiMessageProducer.SetDioPortState(1, 1);
         Assert.Equal("DIO:PORt:STATe 1,1", message.Data);
         AssertMessageFormat(message);
+    }
+
+    [Fact]
+    public void SetDioPortState_WithNegativeChannel_Throws()
+    {
+        Assert.Throws<ArgumentOutOfRangeException>(() => ScpiMessageProducer.SetDioPortState(-1, 1));
+    }
+
+    [Fact]
+    public void SetDioPortState_WithChannelZero_StillReturnsCommand()
+    {
+        var message = ScpiMessageProducer.SetDioPortState(0, 1);
+        Assert.Equal("DIO:PORt:STATe 0,1", message.Data);
+        AssertMessageFormat(message);
+    }
+
+    [Fact]
+    public void SetDioPortState_WithNegativeChannel_ThrowsBeforeCheckingTheValue()
+    {
+        // Both arguments are invalid here. The channel guard runs first, so the caller is
+        // told about `channel` rather than about `value`, matching the sibling setters that
+        // validate the channel before the value.
+        var ex = Assert.Throws<ArgumentOutOfRangeException>(
+            () => ScpiMessageProducer.SetDioPortState(-1, double.NaN));
+
+        Assert.Equal("channel", ex.ParamName);
     }
 
     [Theory]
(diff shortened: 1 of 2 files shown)


tylerkron:
Qodo-clean, CI green — ready for review. (Round 1 verdict on head `f2b9314`: Bugs 0 / Rule violations 0 / Requirement gaps 0, 0 unresolved inline threads; independently re-verified after a settle wait, thread query sanity-checked against #627. Test-merge with current `origin/main` builds with 0 warnings and the full suite is green on net9.0 + net10.0 — 3832 + 3832 + 217. Note for the reviewer: this deliberately rejects input that used to be accepted — a negative channel on the two DIO producers now throws, matching the other nine channel-addressed commands.)

Issue #553, opened 2026-08-20
TextExchangeEngine's stale-line boundary is captured before the send, so a late reply can still be attributed to the wrong exchange


cptkoolbeenz:
Raised during review of [#552](https://github.com/daqifi/daqifi-core/pull/552), where it was investigated and **correctly declined as pre-existing**. Recording it so the limitation is findable rather than living in a closed review thread.

## The window

`TextExchangeEngine.ExecuteAsync` marks the boundary between "already in flight" and "answers to this exchange" like so:

```csharp
staleLineCount = CollectedLineCount();      // boundary captured
...
await setupActionAsync(cancellationToken);  // command actually sent
```

A line arriving **between** those two points is counted as this exchange's answer, even though the command had not gone out yet. The code already documents the intent — *"Anything captured before the setup action has sent anything is a late reply to an EARLIER command … never a response to a command this exchange has yet to send"* — and the boundary delivers that for anything arriving before the capture. The gap is the interval after it.

## Why it matters more since #552

#552 made a blank line meaningful: `GetSystemLogAsync` reads "any line arrived" as "the device answered", because the firmware terminates every `SYSTem:LOG?` dump with a blank line. A stale blank landing in that window would therefore report an empty log for a device that had gone silent.

**This is not a regression, and #552 did not introduce it.** Traced on both sides: on `main`, `EmitEmptyLines = true` already lets such a late blank satisfy `hasReceivedAny`, and the pre-existing `.Where(line => line.Length > 0)` filter reduces it to an empty list, which `SystemLogParser.Parse` and `ThrowIfErrorOnlyResponse` already treat as "no entries, don't throw". Same input, same output, before and after. It applies to **every** caller of the engine, not just the log read — the SD listing's terminator check (#396) is the case the boundary was originally added for.

## I tried a fix and backed it out — worth knowing before someone retries it

I implemented a second boundary captured immediately after `setupActionAsync` returns, used **only** to reject blank lines (a device cannot emit a terminator for a command it has not been sent, so a blank there is necessarily from an earlier exchange; content lines kept the wider boundary so nothing else changed).

It built, and I wrote two engine-level tests against the existing `ReleaseOnStreamAccessMockTransport` harness — a stale blank released at consumer-bind, and a real blank released on send.

[...]

PR #650, opened 2026-08-24
fix(tests): ask the exchange whether it recognised the reply, not the clock

### LINKED ISSUE #634: test(flake): ExecuteTextCommand_WhenTheDeviceAnswersWithOnlyABlankLine_DoesNotWaitOutTheTimeout has an over-tight 2000 ms bound
Found while validating PR #633 (the fix for #632). This is a **separate, unrelated flake** — it is not caused by that diff.

## What happens

`TextExchangeLineFramingTests.ExecuteTextCommand_WhenTheDeviceAnswersWithOnlyABlankLine_DoesNotWaitOutTheTimeout` failed on CI with:

- net9.0 **red**, net10.0 **green**, on the same commit — the same one-TFM-split signature as #632 and PR #601.
- Reported duration: **took 2129ms**, against an assertion bound of **2000 ms**.
- A re-run on the same SHA went **green**, confirming it is timing-dependent rather than a real regression.

## Why the bound looks wrong

The test is guarding against the exchange *waiting out its timeout*, and the failure mode it exists to catch is a **>= 3000 ms** wait. But the assertion is set at **2000 ms**, which leaves very little headroom above the test's own healthy runtime (~2129 ms observed on a loaded CI runner). So the bound is not discriminating between "fast path" and "waited out the timeout" — it is discriminating between "unloaded runner" and "loaded runner".

Raising the bound toward the actual failure threshold (somewhere between the healthy runtime and 3000 ms) would keep the guard meaningful while removing the false positive. Worth checking whether the deterministic-handshake approach used in PR #633 applies here too, rather than tuning a number.

## Notes

- Not fixed. Filed to track it so it does not get lost.
- Please do **not** fix this by deleting or weakening the assertion — like #632, this test guards real text-exchange timeout behavior.
- Related: #632 (same signature, now fixed by #633), #601 and #609 (both thread-pool starvation racing a fixed timeout).

Filed by the autonomous maintenance loop; no code change was made.

tylerkron:
## What was wrong

Two tests in `TextExchangeLineFramingTests` — the ones that check the SCPI text exchange
recognises a blank-line or bare-LF reply instead of sitting out its response timeout — were
failing intermittently on CI. The most recent one reported "the exchange took 2129ms" against a
2000 ms bound: net9.0 red, net10.0 green on the same commit, and green again on a re-run of the
same SHA. Nothing was actually wrong with the exchange on those runs. The tests were deciding
whether the reply had been recognised by timing the whole call, and the exchange's reply wait
loop polls with `await Task.Delay(50)` — so on a thread-pool-starved build agent an exchange that
recognised the reply on its very first poll still takes seconds of wall clock to return, and the
stopwatch reports a fast path as a timed-out one. The bound was measuring the runner, not the
code under test.

## How it was fixed

The exchange now says which branch its wait loop left by, through a test-only host seam
(`OnReplyWaitCompleted(bool sawResponse)`, a no-op on the real device and the third of these
after the two added for #553 and #632), and the tests assert on that instead of on elapsed time.

The thing worth pushing back on is that this removes a timing assertion, so it is worth being
explicit that the guard is not weakened. The failure these tests exist to catch — a reply shape
the line framing cannot see (#538) — leaves the wait loop in its first-response phase, which now
lands as `false` here; so does the #592 regression of a reply that arrived before the loop's
first poll going unnoticed. Both were previously caught only by inference from the clock. What
the stopwatch added on top of that was the machine's load, and that is the only thing dropped.
The silent-device control test gains the mirror assertion (silence must leave the exchange
believing nothing answered) and keeps its existing elapsed-time check, which can only fail late
and so cannot flake this way.

## Verification

Reproduced with a temporary in-process harness that blocks the thread pool with
`ProcessorCount * 8` non-CPU-burning work items — external CPU spinners do not starve the .NET
pool, which is the actual CI mechanism. Harness removed before committing; no external spinners
were needed, so peak extra CPU load was zero.

| | net9.0 | net10.0 |
|---|---|---|
| before | 10 / 10 runs failed | 10 / 10 runs failed |
| after | 0 / 10 | 0 / 10 |

Observed failure times before the fix were 2.1–2.7 s against the 2000 ms bound — the same range
as the CI failure in the issue. In every one of those runs the exchange returned the correct
result; only the stopwatch disagreed. This is a test-side timing sensitivity, not a product bug.

Full suite green on both target frameworks: Core 3827 passed / 0 failed / 2 skipped on net9.0 and
net10.0, Mcp 217 passed. Build clean, 0 warnings.

closes #634

Not merging — for review.

### DIFF
--- a/src/Daqifi.Core.Tests/Device/Internal/ConnectionGuardTests.cs
+++ b/src/Daqifi.Core.Tests/Device/Internal/ConnectionGuardTests.cs
@@ -189,5 +189,6 @@ public class ConnectionGuardTests
         public IDisposable SubscribeConsumerErrors(IMessageConsumer<string> consumer) => throw new NotSupportedException();
         public void OnStaleLineBoundaryCaptured() => throw new NotSupportedException();
         public void OnSendBoundaryCaptured() => throw new NotSupportedException();
+        public void OnReplyWaitCompleted(bool sawResponse) => throw new NotSupportedException();
     }
 }
--- a/src/Daqifi.Core.Tests/Device/TextExchangeLineFramingTests.cs
+++ b/src/Daqifi.Core.Tests/Device/TextExchangeLineFramingTests.cs
@@ -55,12 +55,10 @@ public class TextExchangeLineFramingTests
 
         device.Connect();
 
-        var stopwatch = Stopwatch.StartNew();
         var lines = await device.CallAsync(() => transport.Release());
-        stopwatch.Stop();
 
         Assert.Equal(new[] { "Added test log messages" }, lines);
-        AssertAnsweredPromptly(stopwatch);
+        AssertRecognisedTheReply(device);
 
         device.Disconnect();
     }
@@ -77,11 +75,9 @@ public class TextExchangeLineFramingTests
 
         device.Connect();
 
-        var stopwatch = Stopwatch.StartNew();
         var lines = await device.CallAsync(() => transport.Release());
-        stopwatch.Stop();
 
-        AssertAnsweredPromptly(stopwatch);
+        AssertRecognisedTheReply(device);
 
         // ...and the blank line stays out of the caller's result. It is evidence for the wait
         // loop, not content: every caller of this seam parses lines, and none of them ever saw a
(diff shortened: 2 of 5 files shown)


tylerkron:

Shepherd verification on head `c16f1b7`: Qodo `Bugs (0) / Rule violations (0) / Requirement gaps (0)` anchored to this SHA and never edited since it landed, zero unresolved inline review threads (query sanity-checked against a PR known to have them); CI `build` SUCCESS.

Additionally re-checked against current `main`, which has since gained #647 (root `Directory.Build.props`, `TreatWarningsAsErrors` now on for every project including Daqifi.Mcp). `main` merges into this branch cleanly, and the merged tree builds **0 warnings / 0 errors** under the stricter settings across all four projects on both target frameworks — the new production-side seam introduces no analyzer or nullability warnings. Full suite green on that merged tree: Core 3833 passed / 0 failed / 2 skipped on net9.0 and net10.0, Mcp 217 passed.

The `OnReplyWaitCompleted` seam is shaped exactly like the two that precede it (`OnStaleLineBoundaryCaptured`, `OnSendBoundaryCaptured`): explicit `ITextExchangeHost` implementation delegating to an empty `internal virtual` method, so it is inert on the real device — a single straight-line call after the wait loop that reads a local `bool` and changes no state.

Not merging.

PR #656, opened 2026-08-24
test(consumers): make the parser routing tests actually test the routing

### LINKED ISSUE #653: The composite/protobuf parser tests named for routing behavior assert almost nothing
Found by the **useless-test-pruner** maintenance routine. Filing rather than fixing because the fix is a rewrite, not a deletion — pruning these would remove the appearance of coverage without adding the real thing, and a speculative PR rewriting eight tests is not something a reviewer can evaluate quickly. (An earlier pass of this routine flagged one of these and noted it needed a rewrite; this issue is that note, written down properly.)

## The problem

A cluster of tests in `CompositeMessageParserTests` and `ProtobufMessageParserTests` have names that promise a routing decision — "ShouldUseProtobufParser", "ShouldDetectAsBinary", "ShouldDetectAsText" — but their only assertion is `Assert.True(consumedBytes >= 0)`. That is very nearly a tautology, and more to the point it says nothing whatsoever about which parser ran, which is the entire behavior the test is named for. A change that routed every input to the wrong parser would leave all of them green.

One is a strict tautology: `CompositeMessageParser_ParseMessages_ReturnsDifferentMessageTypes` ends with `Assert.True(binaryMessages.Count() >= 0)`, and `Enumerable.Count()` cannot return a negative number.

Several also guard their only meaningful assertion behind `if (messages.Any())`, so when the parser returns nothing the test asserts nothing at all and still passes.

## Affected tests

`src/Daqifi.Core.Tests/Communication/Consumers/CompositeMessageParserTests.cs`
- `..._WithBinaryData_ShouldUseProtobufParser` (:40)
- `..._WithNullBytes_ShouldDetectAsBinary` (:55)
- `..._WithNullTextParser_ShouldStillWork` (:122)
- `..._WithNullProtobufParser_ShouldStillWork` (:136)
- `..._ReturnsDifferentMessageTypes` (:160, the `Count() >= 0` tautology)
- `..._WithMixedScenarios_ShouldHandleGracefully` (:186)

`src/Daqifi.Core.Tests/Communication/Consumers/ProtobufMessageParserTests.cs`
- `..._WithValidProtobuf_ShouldReturnMessage` (:23-27, conditional assert plus `>= 0`)
- `..._WithNullBytes_ShouldHandleCorrectly` (:72)
- `..._WithMultipleMessages_ShouldParseFirst` (:106)

## Suggested direction

Assert what the parser actually returns for each input: the concrete message type (`ProtobufMessage` vs the text message type), the message count, and the exact `consumedBytes` value — the same style the rest of `ProtobufMessageParserTests` already uses well (see `RecoversFromLeadingGarbage`, `RecoversFrameWrappedInEchoAndPrompt`, which pin exact counts, decoded field values, and exact consumed byte counts). The "should not throw" intent that several of these seem to be reaching for is better said with `Record.Exception(...)` and an `Assert.Null`, which `WithMixedScenarios` already does — its `consumed >= 0` line inside the loop is the redundant part.

Worth confirming while doing this whether each input actually is routed the way the test name claims; if `CompositeMessageParser` does not in fact pick the protobuf parser for some of these inputs, the names are wrong and that is the more interesting finding.

Note: the rest of `ProtobufMessageParserTests` is strong and should be left alone — this is specifically the older, weaker block near the top of each file.

tylerkron:
## What was wrong

Ten tests across `CompositeMessageParserTests` and `ProtobufMessageParserTests` carried the names of routing tests — `ShouldUseProtobufParser`, `ShouldDetectAsBinary`, `WithMultipleMessages_ShouldParseFirst` — but their only assertion was `Assert.True(consumedBytes >= 0)`. Nothing in them said which parser had run, which is the entire behaviour they were named for. A change that routed every buffer to the wrong parser would have left all ten green. One was a strict tautology (`Assert.True(binaryMessages.Count() >= 0)` — `Enumerable.Count()` cannot be negative), and five hid their only real assertion behind `if (messages.Any())`, so they asserted nothing whenever the parser returned empty — which, for the buffers they used, was always. Three of them ran on an **empty** byte array despite being named for valid protobuf. The suite was reporting coverage of the parser routing that the project did not actually have.

## How it was fixed

Each test now pins the real behaviour: which parser was consulted and in what order, how many messages came out, of what concrete type and with what decoded field values, and what `consumedBytes` exactly equals. Two recording parsers write into one shared ordered call log, so a test can assert the routing **order** (`["protobuf", "text"]`) rather than merely that a parser was reached — and can prove the negative, that the second parser was never consulted when the first answered. Every `if (messages.Any())` guard is gone. The buffers that were fake "mock protobuf" bytes (not valid length-delimited frames at all, which is why nothing ever parsed) are now real frames built with `WriteDelimitedTo`.

**Two things a reviewer should push back on if they disagree:**

1. **`WithNullBytes_ShouldDetectAsBinary` was named for the opposite of what the code does.** `"Hello\0World"` is 90.9% printable ASCII and 9.1% null, and `DetectMessageType` weighs printable ratio *before* null density on purpose (the 10% null threshold isn't even met). It routes to **text**. That is deliberate and documented in the classifier — the genuinely binary case it guards against is the 40%-null / 24%-printable `SYSInfoPB?` reply from issue #268 — so I renamed the test to `WithOneNullByteInPrintableData_ShouldDetectAsText` rather than pin a behaviour the code doesn't have. **This is not a bug report**; no parser bug was found. If you'd rather the classifier *did* treat any null byte as binary, that's a separate design change and this test would need to move with it.

2. **`ReturnsCorrectMessageType` is a tenth test, not in issue #653's list of nine.** It sits in the same file with the identical defect (empty buffer, `if (messages.Any())`, `consumedBytes >= 0`). Leaving it would have shipped a file that still contained the exact pattern this PR claims to remove.

### Mutation evidence — every rewritten test was confirmed able to fail

Each mutation was applied to production code, the tests run, then reverted. Verified red:

| Mutation | Tests it turns red |
|---|---|
| M1 — swap the two routing branches in `CompositeMessageParser.ParseMessages` | `WithBinaryData_ShouldUseProtobufParser`, `WithOneNullByteInPrintableData_ShouldDetectAsText`, `WithNullTextParser_UsesDefaultLineParser`, `WithNullProtobufParser_UsesDefaultProtobufParser`, `WhenPreferredParserFindsNothing_FallsBackWithFullBuffer` |
| M2 — `TryParse` adopts `Math.Max(consumedBytes, parserConsumed)` instead of only adopting on a produced message | `WithMixedScenarios` rows *truncated protobuf frame*, *binary with nulls* |
| M3 — `currentIndex += messageLength` (drop the length prefix from the advance) | `WithValidProtobuf_ShouldReturnMessage`, `WithMultipleMessages_ShouldParseAll`, `ReturnsCorrectMessageType`, `ReturnsDifferentMessageTypes`, `WithNullProtobufParser_UsesDefaultProtobufParser` |
| M4 — resync advances by 2 instead of 1 in the invalid-length branch | `WithNullBytes_ResyncsAndPreservesTrailingPrefix` (and only that one) |
| M5 — `break` after the first parsed frame | `WithMultipleMessages_ShouldParseAll` |
| M6 — `TextPrintableRatio` 0.8 → 0.95 | `WithOneNullByteInPrintableData_ShouldDetectAsText` |
| M7 — `LineBasedMessageParser` reports `consumedBytes` without the line ending | `ReturnsDifferentMessageTypes`, `WithNullTextParser_UsesDefaultLineParser`, `WithMixedScenarios` rows *SCPI command*, *SCPI query* |
| M8 — `LineBasedMessageParser` swallows an unterminated tail as a line | `WithMixedScenarios` rows *lone non-printable byte*, *truncated protobuf frame*, *binary with nulls* |

The one input I could **not** make failable is the empty buffer, because `CompositeMessageParser` returns before routing — so I dropped it from the theory rather than keep a row no mutation can kill; the existing `WithEmptyData_ShouldReturnEmpty` already pins that contract exactly. There is a comment in the theory data saying so.

## Verification

Test-only change; no production code is modified. Full suite green on both frameworks: net9.0 Core 3838 passed / 2 skipped, net10.0 Core 3838 passed / 2 skipped, Mcp 217 passed, 0 failed, 0 warnings. Line coverage 88.48%. Bench validation does not apply — nothing here touches the device, and the board was never acquired.

The diff also drops trailing whitespace from four adjacent tests I did not otherwise change, a side effect of rewriting the file.

closes #653

**Not merging — for review.**

### DIFF
--- a/src/Daqifi.Core.Tests/Communication/Consumers/CompositeMessageParserTests.cs
+++ b/src/Daqifi.Core.Tests/Communication/Consumers/CompositeMessageParserTests.cs
@@ -1,5 +1,6 @@
 using Daqifi.Core.Communication.Consumers;
 using Daqifi.Core.Communication.Messages;
+using Google.Protobuf;
 using System.Text;
 
 namespace Daqifi.Core.Tests.Communication.Consumers;
@@ -12,10 +13,10 @@ public class CompositeMessageParserTests
         // Arrange
         var parser = new CompositeMessageParser();
         var textData = Encoding.UTF8.GetBytes("*IDN?\r\nDAQiFi Device v1.0\r\n");
-        
+
         // Act
         var messages = parser.ParseMessages(textData, out var consumedBytes);
-        
+
         // Assert
         Assert.Equal(2, messages.Count());
         Assert.All(messages, msg => Assert.IsType<string>(msg.Data));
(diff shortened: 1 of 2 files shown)


tylerkron:
Qodo-clean, CI green — ready for review. Independently re-verified on head `09c66ee`: Qodo Bugs 0 / Rule violations 0 / Requirement gaps 0 with zero review threads (settle re-check 4 min later unchanged; anti-artifact check on #627 returned its 2 threads), CI `build` success on that exact SHA, `dotnet build --no-incremental -warnaserror` clean at 0 warnings on both TFMs, and the full suite green (Core 3838 net9.0 + 3838 net10.0, Mcp 217, 0 failed). Re-derived three of the mutations myself rather than trusting the table — M1 (swap the routing branches) killed exactly the 5 listed tests, M3 (drop the length prefix from the advance) killed the 5 listed plus 4 pre-existing, and M4 (resync advance of 2) killed exactly `WithNullBytes_ResyncsAndPreservesTrailingPrefix`. Also confirmed the `WithOneNullByteInPrintableData_ShouldDetectAsText` rename against the classifier source: `"Hello\0World"` is 10/11 printable content bytes (0.909 > `TextPrintableRatio` 0.8) so heuristic 1 returns `LikelyText` — and its null ratio is 1/11 = 0.0909, which is *not* above `BinaryNullRatio` 0.1, so the buffer classifies as text under either ordering of the two heuristics. The rename is therefore not papering over a precedence question; a reviewer who wants any null byte to force binary is asking for a separate design change.

Issue #655, opened 2026-08-24
feat(export): the SD-card half of #639 — Core still ships no ISampleSource over SdCardLogSession, and there are now three hand-rolled copies


tylerkron:
**Type**: Feature / API completeness
**Priority**: Medium
**Follows**: #639 (the live half landed in PR #TBD; this is the other half of that ticket)

## What

#639 asked for two things. The live half — a public `ISampleSource` over a device's live stream,
plus a recorder that drives it through `CsvExporter` — is done. The SD-card half is not, and it is
the smaller and more obviously duplicated of the two.

`Daqifi.Core` ships the SD-card parsers that produce `SdCardLogSession` (`Device/SdCard/`, three
formats) and it ships the `CsvExporter` that consumes `ISampleSource`. It still ships no adapter
between them. There are now **three** hand-rolled copies of that adapter outside Core:

```
src/Daqifi.Mcp/SdCardSampleSource.cs                          (internal, 106 lines)
src/Daqifi.Core.Tests/Logging/Export/InMemorySampleSource.cs  (a fixture — stays internal)
```

and, in the example-app repo:

```
daqifi-core-example-app/SdCardSampleSource.cs                 (internal, same shape)
```

The MCP copy and the example-app copy are near-identical: same `DeviceName = "Daqifi"` constant,
same `DigitalChannelName = "DIO"`, same one-column-per-analog-channel-plus-DIO layout, same
`GetSampleCountAsync` returning 0. That is the duplication #481 and #477 were filed for, applied to
the export layer — and it has already happened twice, before `daqifi-desktop` (the obvious third
consumer) has even asked for it.

## Why it matters beyond the duplication

`SdCardSampleSource`'s own remarks document a trap every reimplementation hits independently:

> `SampleCount` is log entries read, `RowCount` is CSV lines written. The exporter collapses every
> consecutive `SampleRow` sharing a timestamp into one line, so a row is a timestamp, not a sample —
> counting samples would over-report by the channel count.

That is a correctness note about `CsvExporter`'s behaviour. It belongs next to the exporter in Core,
not rediscovered per consumer. The live half of #639 had to restate the same rule in
`LiveSampleSource.RowCount`; that is now two places in the tree saying the same thing about the same
exporter, and neither is the one a package consumer can reach.

## How

1. Promote the adapter into `Daqifi.Core.Logging.Export` as a public `SdCardLogSampleSource` (or an
[...]

PR #657, opened 2026-08-24
feat(export): let a live stream be recorded straight to CSV

### LINKED ISSUE #639: feat(export): Core ships CsvExporter but no ISampleSource — every consumer hand-rolls the adapter, and live streams cannot be recorded at all
**Type**: Feature / API completeness
**Priority**: Medium-High

## What
`Daqifi.Core.Logging.Export` ships a public `CsvExporter` that consumes a public `ISampleSource`. Core ships **zero implementations of `ISampleSource`**.

Every implementation that exists is invisible to a package consumer:

```
$ grep -rn "ISampleSource" src --include="*.cs" | grep -v "ISampleSource.cs"
src/Daqifi.Core.Tests/Logging/Export/InMemorySampleSource.cs:5:  internal sealed class InMemorySampleSource : ISampleSource
src/Daqifi.Mcp/SdCardSampleSource.cs:20:               internal sealed class SdCardSampleSource : ISampleSource
```

One is `internal` to the test project. The other is `internal` to `Daqifi.Mcp` — 106 lines whose own doc comment describes it as the missing bridge:

> Adapts a parsed `SdCardLogSession` to `ISampleSource` so Core's `CsvExporter` can turn a downloaded SD-card log into a CSV the agent can read.

So `Daqifi.Core` contains both halves — the SD-card parsers that produce `SdCardLogSession` (`Device/SdCard/`, three formats: protobuf, CSV, JSON) and the exporter that consumes `ISampleSource` — and ships no connection between them. DAQiFi's own MCP server had to write that connection, and marked it `internal`, so nobody else can use it.

There is a second, larger gap on the live side. `ILiveSampleSource` (`Device/ILiveSampleSource.cs`) exposes `IAsyncEnumerable<LiveSample> StreamSamplesAsync(...)`, and its own remarks draw the distinction:

> Not to be confused with `Logging.Export.ISampleSource`, which is the offline side ... This one is the live device.

Nothing adapts live to offline. A consumer who wants the most basic DAQ workflow — *stream from the device and record it to a file* — gets an `IAsyncEnumerable<LiveSample>` from one namespace, an exporter in another, and no supported way to join them.

## Why
This is the exact pattern three existing tickets were filed for, applied to the export layer:

- **#501** — "engineering-unit scaling — Core parses unit and range then drops them; **desktop hand-rolls the feature**"
- **#481** — "the per-configuration sample-rate cap **lives only in `Daqifi.Mcp`** — move it into Core"
- **#477** — "move the bootloader no-op streaming device into Core — **both apps carry a duplicate shim**"

`SdCardSampleSource` is the same shape: functionality that belongs in Core, living in a downstream consumer as an internal class, waiting to be duplicated by the next consumer. `daqifi-desktop` is the obvious next one — it is the GUI over this SDK and "export my log to CSV" is table stakes for it.

The cost is not only duplication. Getting `ISampleSource` right is genuinely fiddly, and `SdCardSampleSource`'s own remarks document a trap that every reimplementation will hit independently:

> `SampleCount` is log entries read, `RowCount` is CSV lines written. The exporter collapses every consecutive `SampleRow` sharing a timestamp into one line, so a row is a timestamp, not a sample — counting samples would over-report by the channel count.

That is a correctness note that belongs in Core next to the exporter, not rediscovered per consumer.

## How
1. **Promote the SD-card adapter into Core.** Move `SdCardSampleSource` from `Daqifi.Mcp` to `Daqifi.Core.Logging.Export` as a public `SdCardLogSampleSource` (or a `SdCardLogSession.AsSampleSource()` extension). `Daqifi.Mcp` deletes its copy and calls the Core one — the same disposition #481 and #477 took.
2. **Add a live recorder.** A `LiveSampleRecorder` (or `ILiveSampleSource.RecordToCsvAsync(path, ct)`) that consumes `StreamSamplesAsync` and writes through `CsvExporter` as samples arrive. This is the piece with no existing implementation anywhere, and it needs decisions Core is the right place to make:
   - `ISampleSource.GetSampleCountAsync` is documented as "may return 0 if the count is unavailable" — a live stream is exactly that case, so progress reporting must degrade rather than divide by zero.
   - `StreamSamples` is contractually "ordered by `TimestampTicks` ascending"; confirm the live path preserves that across the rollover handling in `TimestampProcessor`.
   - `DroppedLiveSampleCount` should be surfaced in the recording result — a recording with silent drops is worse than a failed one.
3. Keep `InMemorySampleSource` internal to tests; it is a fixture, not API.

Step 1 is small and unblocks desktop immediately. Step 2 is the feature.

## Success criteria
- [ ] A public, supported `ISampleSource` over `SdCardLogSession` in `Daqifi.Core`, covering all three SD log formats.
- [ ] `Daqifi.Mcp/SdCardSampleSource.cs` deleted, with the MCP tools calling the Core implementation and its tests still green.
- [ ] A documented path from `ILiveSampleSource` to `CsvExporter` that records while streaming, without buffering the whole session in memory.
- [ ] Dropped-sample count surfaced to the caller of a live recording.
- [ ] `README.md` gains a "record a live stream to CSV" recipe alongside the existing streaming snippet.

tylerkron:
Core shipped both halves of a recording and nothing that joined them. `ILiveSampleSource` handed you an `IAsyncEnumerable<LiveSample>` in one namespace; `CsvExporter` asked for an `ISampleSource` in another; there was no supported way to connect the two. So the most basic DAQ workflow there is — stream from the device and record it to a file — had no answer in the package. Anyone who wanted it had to write the adapter themselves, and the adapter is fiddly enough that DAQiFi's own MCP server, the example app, and the test project each ended up with their own copy.

This adds the join. `LiveSampleSource` is a public `ISampleSource` over a live stream that yields rows as frames decode rather than buffering the session, so a recording's memory does not grow with its length. `IStreamingDevice.RecordLiveSamplesToCsvAsync(writer, options, duration, bufferCapacity, ct)` drives it through `CsvExporter` and returns a `LiveCsvRecordingResult` counting what reached the file and what did not — dropped, unmapped, non-monotonic. A recording with silent gaps is worse than one that failed, so nothing is swallowed.

**Purely additive.** Two new types, one extension method, one README recipe. No existing public signature moved — the breaking-change question from #612/#616 is left to you. ADR 0002's source-compatibility promise only constrains how `LiveCsvRecordingResult` may grow later (positional appends), which is what it permits.

### What a reviewer should push back on

**The extension hangs off `IStreamingDevice` but needs `ILiveSampleSource` too**, so it does a runtime `is not ILiveSampleSource` check and throws `ArgumentException`. The recorder needs channels and identity from one interface and samples from the other, and neither implies the other (`DaqifiStreamingDevice` implements both). The alternatives were passing the same object twice, or a generic constraint that fails inference for anyone holding an `IStreamingDevice`. The runtime check has repo precedent in `DaqifiAgent.RequireLiveSamples`. It is the least-bad of three imperfect options and worth a second opinion.

**`duration` is the clean stop; `cancellationToken` is the abort.** The window arrives as a cancellation (the read it interrupts is unbounded) and is absorbed *inside* the private iterator, not at the exporter — that is what makes a timed recording end with a complete file, because the exporter sees the enumeration finish normally and flushes its last accumulated line. Letting the OCE reach the exporter silently discards that row. There is deliberately **no** graceful "stop signal" for a user-driven Stop button yet; today that means `duration`, or cancel and accept an abort.

**`GetSampleCountAsync` returns 0** — exactly the "count unavailable" case the interface documents, and what makes `CsvExporter` skip percentage progress instead of dividing by an unknown total. Consequently there is no `IProgress<int>` parameter: a percentage of an unbounded stream is meaningless, and `RowCount`/`SampleCount` are the progress.

**`RowCount` is timestamps, not samples.** The exporter collapses consecutive rows sharing a timestamp into one line, so counting samples would over-report by the channel count. That correctness note now lives in Core next to the exporter instead of being rediscovered per consumer.

Also: values are written as `ScaledValue` (a user who configured a transducer asked for PSI, not the volts underneath); samples match channels by `(Type, ChannelNumber)` rather than object identity, because a device rebuilds its channel objects on status repopulation; blank device name/serial normalize to `"unknown"`, the same disposition as #627.

### Verification

Solution builds with `-warnaserror` under the root `Directory.Build.props` from #647: 0 warnings. Full suite green — 3878 Core tests on net9.0 and net10.0, 217 MCP tests, 0 failures. 45 new tests across `LiveSampleSourceTests` (24) and `LiveCsvRecordingTests` (21), weighted to edge cases: null/blank/duplicate channels, type-vs-number collision, empty stream, unmapped samples, backwards timestamps, three flavours of cancellation, no-enabled-channels, writer flush/non-dispose, buffer-capacity forwarding.

**Every test was proved able to fail** via a 19-mutation matrix, each mutation reverted after its run. Which mutation covered which test:

| Mutation to production code | Tests turned red |
|---|---|
| M1 `RowCount++` made unconditional | 4 |
| M2 unmapped samples not counted | 3 |
| M3 `ScaledValue` → `Value` | 1 |
| M4 blank serial kept instead of normalized | 2 |
| M5 duplicate channels kept | 1 |
| M6 non-monotonic never counted | 1 |
| M7 token not forwarded to the wrapped stream | 3 |
| M8 null channel entry allowed | 1 |
| M9b duration OCE rethrown instead of absorbed | 8 |
| M10 dropped reported absolutely, not as a delta | 1 |
| M11 disabled channels included | 3 |
| M12 `bufferCapacity` validation dropped | 2 |
| M13 `duration` validation dropped | 2 |
| M14 writer not flushed | 1 |
| M15 `bufferCapacity` not forwarded | 1 |
| M16b caller-cancel clause dropped from the catch filter | 1 |
| M17 up-front cancellation check dropped | **0** — see below |

M16 initially killed nothing: the `&& !cancellationToken.IsCancellationRequested` half of the filter only matters when *both* tokens are cancelled, which no test reached. Rather than delete a defensible guard or leave it unpinned, a test was added that gates the scripted stream on a `TaskCompletionSource` so it decides exactly which tokens are cancelled when the enumeration throws.

M17 is killed by nothing, and is kept deliberately: `CsvExporter.ExportAsync` happens to make the same check, so no test can distinguish it — but the recorder's contract should not depend on a collaborator's internal ordering. Flagging it rather than quietly leaving it.

**Bench-validated on real hardware** (fw 3.7.2, sn 9090539562006014104), via a scratch console app project-referencing this branch, since the example CLI has no live-CSV command:
- **100 Hz / 5 s:** `rows=395 samples=790 dropped=0 unmapped=0 nonMonotonic=0`. `samples == 2 * rows` exactly — the per-frame collapse holds on hardware. CSV timestamps exactly 10.000 ms apart across all 395 rows.
- **1000 Hz / 4 s with `bufferCapacity: 1`** (deliberately starved): `rows=3006 samples=3025 dropped=3323`. This is the "dropped-sample count surfaced to the caller" criterion demonstrated on hardware — the losses are reported, not silent.

All runs non-destructive: connect, enable channels, stream, record, stop, disconnect.

### Deferred

This is step 2 of #639. Step 1 — promoting `Daqifi.Mcp`'s internal `SdCardSampleSource` into Core — is filed as **#655** and deliberately not here: it is a cross-project *move* (delete, re-point, keep MCP's tests green), and mixing that with new public API makes one PR a reviewer has to evaluate in two modes. #655 also records new evidence the issue did not have: the example app carries a **third** hand-rolled copy of the same adapter.

closes #639

---

**Not merging — for review.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ More examples at [daqifi.com](https://daqifi.com).
 | **One-line connect** | `DaqifiDeviceFactory.ConnectTcpAsync(...)` wraps transport setup and device init; retries are opt-in via `DeviceConnectionOptions` |
 | **Real-time streaming** | Per-channel `IChannel.SampleReceived` events with decoded, scaled values — or subscribe to the raw protobuf frame directly; no polling loops to write |
 | **Acquisition health** | Attach `AcquisitionStatistics` to a stream and read back the rate you are really getting, per-channel jitter, value range, and how far behind the device's clock the host is |
+| **Record to CSV** | `device.RecordLiveSamplesToCsvAsync(writer)` writes a live stream to CSV as it arrives — no buffering the session in memory — and reports what reached the file and what was dropped |
 | **Digital I/O** | Set any DIO pin as input or output and drive outputs high/low; inputs stream alongside analog data |
 | **PWM outputs** | Drive PWM on capable DIO pins with per-channel duty cycle and a shared, device-wide frequency |
 | **SD card operations** | List, download, delete, format, and start/stop SD logging over USB / serial |
@@ -200,6 +201,36 @@ samples went missing; the two disagreeing means the device's own clock is not ke
 it is `MeasuredSampleRateHz` that describes what your application actually received. `Reset()` starts
 a fresh window mid-session, and `stats.Record(sample)` feeds one by hand from `StreamSamplesAsync`
 instead of attaching.
+
+### Record a live stream to CSV
+
+Streaming and exporting used to be two halves with nothing between them. `RecordLiveSamplesToCsvAsync`
+joins them: it writes rows through `CsvExporter` as frames decode, so the recording's memory does not
+grow with its length, and it hands back what reached the file and what did not.
+
+```csharp
+using Daqifi.Core.Logging.Export;
+
+device.StreamingFrequency = 100;
+device.StartStreaming();
+
+await using var writer = new StreamWriter("run.csv");
+var result = await device.RecordLiveSamplesToCsvAsync(writer, duration: TimeSpan.FromSeconds(30));
+
+device.StopStreaming();
+
+Console.WriteLine($"{result.RowCount} rows from {result.SampleCount} samples");
+if (result.DroppedSampleCount > 0)
+{
+    Console.WriteLine($"{result.DroppedSampleCount} samples dropped — raise bufferCapacity or lower the rate");
+}
+```
+
+The columns are the channels that were enabled when the call started, in device order. `duration`
+elapsing is a clean finish — the last frame is written and the result comes back; cancelling the
+`CancellationToken` is an abort and throws, so a recording cut short is never mistaken for a complete
+one. Need the rows somewhere other than a `TextWriter`? Build a `LiveSampleSource` over
+`StreamSamplesAsync` and hand it to `CsvExporter` (or any other `ISampleSource` consumer) yourself.
 
 ### Digital output
 
--- a/src/Daqifi.Core.Tests/Logging/Export/LiveCsvRecordingTests.cs
+++ b/src/Daqifi.Core.Tests/Logging/Export/LiveCsvRecordingTests.cs
@@ -1,0 +1,514 @@
+using System.Globalization;
+using System.Runtime.CompilerServices;
+using System.Text;
+using Daqifi.Core.Channel;
+using Daqifi.Core.Communication.Messages;
+using Daqifi.Core.Device;
+using Daqifi.Core.Logging.Export;
+using Daqifi.Core.Tests.Firmware;
+using ChannelFactory = System.Threading.Channels.Channel;
+
+namespace Daqifi.Core.Tests.Logging.Export;
+
+/// <summary>
+/// Covers <see cref="LiveCsvRecordingExtensions.RecordLiveSamplesToCsvAsync"/> — the join between a
+/// device's live stream and <see cref="CsvExporter"/> that issue #639 was filed for.
+/// </summary>
+/// <remarks>
+/// Most cases drive a device whose live stream is scripted (<see cref="ScriptedLiveDevice"/>), so
+/// what arrives and when is exact rather than timed. The last test drives the real decode path of a
+/// real <see cref="DaqifiStreamingDevice"/> instead, because the argument for this design is that it
+/// records an actual live stream — a scripted one cannot show that the timestamps a frame produces
+/// collapse into one CSV line, or that they come out ascending.
+/// </remarks>
+public class LiveCsvRecordingTests
+{
+    /// <summary>
+    /// Upper bound on every await here. A regression in the stop path is "the recording never
+    /// returns", which would otherwise park the run rather than fail it.
+    /// </summary>
+    private static readonly TimeSpan Timeout = TimeSpan.FromSeconds(15);
+
+    private static readonly DateTime T0 = new(2024, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+    // ── Argument validation ─────────────────────────────────────────────────
+
+    [Fact]
+    public async Task RecordLiveSamplesToCsvAsync_NullDevice_Throws()
+    {
+        var ex = await Assert.ThrowsAsync<ArgumentNullException>(
+            () => ((IStreamingDevice)null!).RecordLiveSamplesToCsvAsync(new StringWriter()));
+
+        Assert.Equal("device", ex.ParamName);
+    }
+
+    [Fact]
+    public async Task RecordLiveSamplesToCsvAsync_NullWriter_Throws()
+    {
+        using var device = CreateScripted();
+
+        var ex = await Assert.ThrowsAsync<ArgumentNullException>(
+            () => device.RecordLiveSamplesToCsvAsync(null!));
+
+        Assert.Equal("writer", ex.ParamName);
+    }
+
+    [Fact]
+    public async Task RecordLiveSamplesToCsvAsync_DeviceThatCannotSupplyLiveSamples_Throws()
+    {
+        // A streaming device is not automatically a live-sample source; the recording has nothing
+        // to read and should say so at the call rather than write an empty file.
+        var device = new FakeStreamingDevice("no-live-samples");
+
+        var ex = await Assert.ThrowsAsync<ArgumentException>(
+            () => device.RecordLiveSamplesToCsvAsync(new StringWriter()));
+
+        Assert.Equal("device", ex.ParamName);
+        Assert.Contains(nameof(ILiveSampleSource), ex.Message);
+    }
+
+    [Theory]
+    [InlineData(0)]
+    [InlineData(-1)]
+    public async Task RecordLiveSamplesToCsvAsync_NonPositiveDuration_Throws(int seconds)
+    {
+        using var device = CreateScripted();
+
+        var ex = await Assert.ThrowsAsync<ArgumentOutOfRangeException>(
+            () => device.RecordLiveSamplesToCsvAsync(
+                new StringWriter(), duration: TimeSpan.FromSeconds(seconds)));
+
+        Assert.Equal("duration", ex.ParamName);
+    }
+
+    [Theory]
+    [InlineData(0)]
+    [InlineData(-1)]
+    public async Task RecordLiveSamplesToCsvAsync_BufferCapacityBelowOne_ThrowsAtTheCall(int capacity)
+    {
+        // The live stream defers this check to the first MoveNextAsync, which happens inside the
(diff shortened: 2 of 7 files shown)


tylerkron:
Qodo round 2 was right on both counts, so both are fixed in 3958ddb rather than rebutted.

**Drop count misattributed (High).** Real, and it broke this method's own contract. `DroppedLiveSampleCount` is device-wide and cumulative across every enumeration; with no enabled channels `CsvExporter` returns before enumerating anything, yet the recording still took a delta over that shared counter — so a concurrent live consumer's drops came back as this recording's, contradicting the documented all-zeros result. The delta is now taken only when there was something to record, and the `DroppedSampleCount` XML doc states plainly that with a second consumer running the number is a delta over a shared counter rather than a per-recording one. Qodo's preferred fix — a recording-scoped counter — would mean new device API for a case a single-consumer caller never hits, so I took the minimal one and documented the residual honestly. New test `..._NoEnabledChannels_DoesNotChargeAnotherConsumersDrops` pins it (the fake now grows the counter between the two reads, so it fails without the fix).

**Fixed-delay test flakiness (Medium).** Also real, and the same shape #633 removed elsewhere today. `LiveSampleStream` now publishes an internal `LiveSubscriptionCount`, incremented once every sample handler is attached, so the test waits for the subscription itself instead of a guessed 400 ms. Internal-only, behind the existing `InternalsVisibleTo`.

Bench validation re-run fresh on hardware this round (fw 3.7.2, sn 9090539562006014104), not inherited:
- 100 Hz / 5 s: `rows=395 samples=790 dropped=0 unmapped=0 nonMonotonic=0`, `samples == 2 * rows`, CSV timestamps 10.000 ms apart across all 395 rows.
- 1000 Hz / 4 s at `bufferCapacity: 1` (deliberately starved): `rows=3162 samples=3170 dropped=3175` — losses surfaced to the caller, and visible in the file as empty cells rather than silent gaps.

Also re-verified independently this round: the change is still purely additive against current `main` (`f015149`, unmoved — 0 deletions in the diff, no existing public signature touched), and two more of the mutation matrix reproduce — M11 (disabled channels included) → 3 red, exactly as claimed; M7 (token not forwarded) → 4 red, the 3 claimed plus `LiveSampleSourceTests.StreamSamples_Cancellation_EndsTheEnumeration`, which my slightly broader form of the mutation also kills.

Full suite green on the new head: 3879 Core on net9.0 and net10.0, 217 MCP, 0 failures; solution builds `-warnaserror` clean.

tylerkron:
Qodo round 3 found one new bug and it is valid, so it is fixed in `ec52604` rather than rebutted.

**Non-monotonic test deadline (Reliability).** The subscription-wait loop added in `3958ddb` — itself the fix for round 2's fixed-delay race — gated its timeout on `DateTime.UtcNow`. That is a wall clock, so an NTP step mid-wait either cuts the wait short (losing exactly the leading frames the loop exists to protect, reintroducing the race under a different cause) or stretches it past the intended bound. Now a `Stopwatch`. Qodo's cited precedent holds: #602 replaced a `UtcNow` deadline for the same reason, and `Stopwatch` is what every other wait loop in this suite already uses — the remaining `DateTime.UtcNow` hits in the test project are sample timestamps, not deadlines.

Test-only change; no production code touched, so the hardware bench result on `3958ddb` still stands unchanged.

Full suite on the new head: **3879 Core passed on net9.0 and net10.0, 217 MCP, 0 failures**, solution `-warnaserror` clean. Also test-merged current `origin/main` (`3e77696`, including #648 and #649) into a scratch branch: merges cleanly and the suite is green there too (3884 per TFM).

Round 2's two findings remain fixed and are shown as resolved.

tylerkron:

Round 4 on `ec52604` reports `🐞 Bugs (0) 📘 Rule violations (0) 📎 Requirement gaps (0) 🎨 UX issues (0) 🔗 Cross-repo conflicts (0) 📜 Skill insights (0)`, with all three findings raised across rounds 2-3 shown resolved and every evidence link anchored to `ec52604`. All 3 inline review threads resolved, 0 unresolved. Re-checked both surfaces after a settle window — unchanged. CI `build` green on this head.

Test-merged current `origin/main` (`3e77696`, i.e. #648 and #649 merged) into a scratch branch: merges cleanly, builds `-warnaserror` clean, suite green at 3884 Core per TFM on net9.0 and net10.0 plus 217 MCP. Still purely additive against current main — 1484 insertions, 0 deletions.

Bench validation was fresh against the real board, not inherited: 100 Hz / 5 s gave rows=395, samples=790, **dropped=0**, CSV timestamps exactly 10.000 ms apart; a deliberately starved 1000 Hz / 4 s at `bufferCapacity: 1` gave **dropped=3175**, surfacing the loss to the caller. The only change since that run is a 5-line test-only fix (`Stopwatch` for `DateTime.UtcNow` in one wait loop), so those numbers still stand.

Issue #629, opened 2026-08-23
ScpiMessageProducer: SetDioPortState neither rejects non-finite doubles nor formats them invariantly, unlike its three sibling double setters


tylerkron:
Noticed while unifying the non-finite guard in `ScpiMessageProducer` (PR to follow). Filing separately because fixing it is a behavior change to a public method, not a refactor.

This is a **different defect in the same method** as #624, which covers only the missing negative-*channel* guard. This one is about the `double value` argument.

## The asymmetry

Four public `ScpiMessageProducer` methods take a `double` and render it into the command text. Three of them do two things before doing so; one does neither.

| Method | Rejects non-finite? | Formats invariant? |
| --- | --- | --- |
| `SetAnalogOutputVoltage` | yes | yes |
| `SetAdcCalibrationSlope` | yes | yes |
| `SetAdcCalibrationOffset` | yes | yes |
| **`SetDioPortState`** | **no** | **no** |

The three conforming methods look like this:

```csharp
if (!double.IsFinite(voltage)) { throw new ArgumentOutOfRangeException(...); }
return new ScpiMessage($"SOURce:VOLTage:LEVel {channel},{voltage.ToString(CultureInfo.InvariantCulture)}");
```

`SetDioPortState` is the whole method:

```csharp
public static IOutboundMessage<string> SetDioPortState(int channel, double value)
{
    return new ScpiMessage($"DIO:PORt:STATe {channel},{value}");
}
```

## Why it matters

Two concrete ways this produces a malformed frame:

1. **Non-finite values reach the wire.** `SetDioPortState(1, double.NaN)` yields `DIO:PORt:STATe 1,NaN`. Its three siblings throw on the same input. `double.PositiveInfinity` is worse — `NumberFormatInfo.PositiveInfinitySymbol` is culture-dependent and renders as `∞` under many cultures, so a non-ASCII character reaches a SCPI parser.

[...]

PR #660, opened 2026-08-24
fix(scpi): reject an out-of-range direction on SetDioPortDirection

### LINKED ISSUE #651: ScpiMessageProducer.SetDioPortDirection accepts any int direction though it documents only 0 and 1
Split out of #624 rather than folded into it, since #624's scope was the negative-channel guard on the two DIO producers and this is a different argument.

## The gap

`ScpiMessageProducer.SetDioPortDirection(int channel, int direction)` documents its second argument as a two-value enumeration:

```csharp
/// <param name="direction">The direction value (0 = input, 1 = output).</param>
```

but accepts any `int`. `SetDioPortDirection(0, 7)` returns a well-formed message whose payload is `DIO:PORt:DIRection 0,7`, and that frame goes to the wire.

Its PWM neighbours in the same class do range-check the equivalent argument:

| Producer | Second argument | Checked? |
| --- | --- | --- |
| `SetPwmChannelFrequency` | `int frequencyHz` | yes — `<= 0` throws "Frequency must be positive." |
| `SetPwmChannelDutyCycle` | `int dutyCyclePercent` | yes — outside `0..100` throws |
| **`SetDioPortDirection`** | **`int direction` (documented 0 or 1)** | **no** |

So the class is inconsistent about whether a documented-range argument is enforced: the two PWM setters enforce theirs, the DIO direction setter does not.

## Why it is worth deciding

The device-level API is already stricter than the producer. `ChannelControlOperations.SetDioDirection` rejects anything that is not `ChannelDirection.Input` or `ChannelDirection.Output` before it calls the producer, so the device path is safe. But `ScpiMessageProducer` is public, and a caller reaching it directly gets no such protection — the same shape of gap #624 just closed for `channel`.

## Related, and deliberately listed separately

`SetDioPortState(int channel, double value)` documents `0 = low, 1 = high` and does not range-check `value` either. That one was raised on #645 (Qodo suggested "validate range (only allow 0/1) in addition to finiteness") and declined there as beyond that PR's stated scope — the `double` type on a two-state argument is itself the anomaly, which #629 records as a breaking-API-change discussion for a major version. Noting it here so the two are decided together rather than one at a time, but the `direction` case is the simpler one: it is already an `int`, so a range check is not a breaking type change, only a behavior change.

## Options

1. Range-check `direction` (`is < 0 or > 1` throws `ArgumentOutOfRangeException`), matching `SetPwmChannelDutyCycle`. Behavior change; nothing in-repo is affected, since the only production caller passes a bool-derived `0`/`1`.
2. Change the parameter type to `ChannelDirection`. Cleaner contract, but a breaking API change belonging with the `SetDioPortState(double)` discussion in #629.
3. Leave it and document the argument as unvalidated. Only defensible if some caller needs to emit an out-of-range direction, which seems unlikely given the firmware only understands 0 and 1.

Recommend option 1, as its own change so the behavior shift is visible.

Found while implementing #624 (PR #649). Cross-references: #624, #629, #645.

tylerkron:
## What was wrong

`ScpiMessageProducer.SetDioPortDirection(int channel, int direction)` documents its second argument as a two-value enumeration — "0 = input, 1 = output" — but accepted any `int` at all. `SetDioPortDirection(0, 7)` handed back a perfectly well-formed message whose payload was `DIO:PORt:DIRection 0,7`, and that frame went out on the wire to firmware that understands only 0 and 1. A caller who got the argument order wrong, or passed a raw pin bitmask, got no complaint from the library — just a nonsense command sent to the board.

The class was also inconsistent with itself: the two PWM setters right beside it do enforce their documented ranges (`SetPwmChannelFrequency` rejects `<= 0`, `SetPwmChannelDutyCycle` rejects outside `0..100`), and `#649` just brought both DIO producers in line on the *channel* argument. `direction` was the remaining hole.

## How it was fixed

`SetDioPortDirection` now range-checks `direction` the same way its PWM neighbours check theirs — an inline `is < 0 or > 1` test right after the channel guard, throwing `ArgumentOutOfRangeException` named for the parameter and carrying the offending value. **This is a behavior change: input that used to produce a message now throws.** That is the point of the ticket, but it is the thing to weigh — anything outside 0/1 was previously accepted silently.

Nothing in the repo is affected. The only production caller, `ChannelControlOperations.SetDioDirection`, already rejects anything that is not `ChannelDirection.Input`/`Output` and passes a bool-derived `0` or `1`. So the change is only reachable by an external caller using the public producer directly — exactly the misuse it is meant to catch. The sibling `SetDioPortState(int, double)` value range is deliberately **not** touched here; its `double`-typed two-state argument is a breaking-API question tracked on #629, and #651 asks for these to stay separate.

## How it was verified

Before changing anything, a throwaway probe recorded what the method actually did on unchanged code. `SetDioPortDirection(0, direction)` for `direction` of `-1`, `2`, `7`, and `int.MaxValue` each returned a message rather than throwing, with payloads `DIO:PORt:DIRection 0,-1`, `0,2`, `0,7`, `0,2147483647`. The same probe pinned the reference shape of the two already-guarded PWM siblings (`ParamName`/`ActualValue`/message) so the new assertions compare against proven behavior rather than a literal copied out of the source.

The five new/changed assertions were then confirmed to **fail against unchanged code** and pass after the fix. The no-regression tests — both documented directions `0` and `1`, and channel `0` — passed in both states, so valid input is byte-identical before and after.

A new `EveryValueRangedProducer_RejectsOutOfRangeArgumentIdentically` test is the companion to the existing channel-uniformity test, on the other axis: it pins the observable throw shape across all three range-checked value arguments, so a fourth added without a guard fails the suite instead of shipping.

Clean `dotnet build --no-incremental`: 0 warnings, 0 errors under repo-wide `TreatWarningsAsErrors`. Full suite green on **net9.0 and net10.0** (3846 passed / 0 failed / 2 pre-existing skips per TFM; Daqifi.Mcp.Tests 217 passed).

closes #651

**Not merging — for review.**

### DIFF
--- a/src/Daqifi.Core.Tests/Communication/Producers/ScpiMessageProducerTests.cs
+++ b/src/Daqifi.Core.Tests/Communication/Producers/ScpiMessageProducerTests.cs
@@ -196,6 +196,47 @@ public class ScpiMessageProducerTests
         AssertMessageFormat(message);
     }
 
+    // `direction` is documented as a two-value enumeration (0 = input, 1 = output) and the
+    // firmware understands nothing else, so anything outside that is caller error rather
+    // than a frame to put on the wire.
+    [Theory]
+    [InlineData(-1)]
+    [InlineData(2)]
+    [InlineData(7)]
+    [InlineData(int.MaxValue)]
+    public void SetDioPortDirection_WithOutOfRangeDirection_Throws(int direction)
+    {
+        var ex = Assert.Throws<ArgumentOutOfRangeException>(
+            () => ScpiMessageProducer.SetDioPortDirection(0, direction));
+
+        Assert.Equal("direction", ex.ParamName);
+        Assert.Equal(direction, ex.ActualValue);
+        Assert.StartsWith("Direction must be 0 (input) or 1 (output).", ex.Message, StringComparison.Ordinal);
+    }
+
+    // Both documented values must survive the guard unchanged, byte for byte.
+    [Theory]
+    [InlineData(0)]
+    [InlineData(1)]
+    public void SetDioPortDirection_WithDocumentedDirection_StillReturnsCommand(int direction)
+    {
+        var message = ScpiMessageProducer.SetDioPortDirection(3, direction);
+        Assert.Equal($"DIO:PORt:DIRection 3,{direction}", message.Data);
+        AssertMessageFormat(message);
+    }
+
+    [Fact]
+    public void SetDioPortDirection_WithNegativeChannel_ThrowsBeforeCheckingTheDirection()
+    {
+        // Both arguments are invalid here. The channel guard runs first, so the caller is
+        // told about `channel` rather than about `direction`, matching SetDioPortState and
+        // the PWM setters, which all validate the channel before the value.
+        var ex = Assert.Throws<ArgumentOutOfRangeException>(
+            () => ScpiMessageProducer.SetDioPortDirection(-1, 7));
+
+        Assert.Equal("channel", ex.ParamName);
+    }
+
     [Fact]
     public void SetDioPortState_ReturnsCorrectCommand()
     {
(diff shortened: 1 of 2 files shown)


tylerkron:
**Conceded and fixed in `ee17745`.** Qodo is right on the substance: the census comment claimed to be the whole population of inline `ArgumentOutOfRangeException` sites, and `SetStreamInterface` was missing from it. Its `Enum.IsDefined` check is the same rule expressed over an enum's range and throws in the identical shape (`ParamName=streamInterface`, actual value carried, `"Unknown stream interface."`), so by the census's own stated criterion it belonged. It is now the ninth entry, and the list matches a grep of that `throw` across the class minus the two shared helpers (`ValidateChannel`, `RequireFinite`).

On the secondary suggestion — replacing the count with a mechanism that cannot silently omit a producer — I've kept the literal count deliberately, and made the comment honest about what it is. There is no signature-level signal that a producer has an inline range guard (the arguments are `int`, `long`, `double` and an enum, with no shared attribute or helper to reflect over), so a reflective census would have to invoke every public static method with a guessed-invalid value, which is more brittle than the table it replaced. The literal count is also the established convention in this file: `EveryChannelAddressedProducer_RejectsNegativeChannelIdentically` uses exactly the same hardcoded-list-plus-`Assert.Equal(11, ...)` shape. It is documented as a tripwire, not a proof of completeness.

Full suite green on net9.0 and net10.0 (3846 passed / 2 pre-existing skips per TFM) with 0 warnings under `TreatWarningsAsErrors`.

tylerkron:
Qodo-clean, CI green — ready for review. (Head `ee17745`: Bugs 0 / Rule violations 0, re-confirmed after settle; the one inline thread is resolved and outdated. Suite green net9.0+net10.0, 0 warnings under `TreatWarningsAsErrors`; CI `build` pass on the same SHA.)

PR #665, opened 2026-08-24
test(sdcard): direct tests for the text log line reader's byte counter and lease discipline

tylerkron:
The **coverage-gap** maintenance routine produced this. It is safe because it adds one new test file and changes no production code at all — `git diff` against main touches nothing under `src/Daqifi.Core/`.

## What was wrong

`SdCardTextLineReader` is the line pump that every line of a downloaded CSV or JSON log passes through on its way to a parser, and it had no tests of its own. Each line it hands over carries a running byte count, and that count is what a caller turns into a download progress bar — so if it drifts, a user watching a large log parse sees a bar that stalls, jumps, or finishes early, and nothing in the suite notices.

Nothing noticed because the CSV and JSON parser fixtures reach this class only along one narrow path: a small, seekable `MemoryStream` positioned at byte zero, enumerated once, asserted on by the samples that come out the far end. They never read the byte count at all, and they never take the forward-only branch. That left the interesting half of the class unpinned:

- the byte count means two different things depending on the source. A seekable stream reports its real `Position`; a forward-only one reports an estimate built from line lengths. Either branch could be deleted and every existing test would still pass.
- blank lines are dropped from the output but still counted toward that estimate. Move the counter one line down, inside the blank check, and a log with blank lines silently under-reports its progress forever.
- reading starts at the stream's *current* position, not at zero — the parsers hand over a stream they have already read a configuration prefix from, so a rewrite that seeks to zero would re-feed the header as data.
- the source overload holds a lease on the source's single read cursor. It has to give that cursor back when the consumer stops early, not only when the file runs out; a leaked lease makes every later read of the same log throw "already being read".

## How it was fixed

Adds `SdCardTextLineReaderTests.cs` — 13 cases over the two `ReadLinesAsync` overloads and `ToAsyncEnumerable`. Tests only, no production change; the reader behaves correctly on every edge probed, so no bug was found and no issue opened.

The byte-count cases are built so the two branches produce visibly different numbers. A 20-byte payload fits in one `StreamReader` buffer, so a seekable stream is already at EOF when the first line arrives and all three lines report `20`; the same payload read forward-only reports `6, 12, 20`. Neither branch can be substituted for the other without a failure.

**Mutation evidence** — eleven mutations of `SdCardTextLineReader.cs`, each applied from a pristine copy and reverted after. All 13 cases die in at least one:

| Mutation | Kills |
|---|---|
| A — move the counter inside the blank-line check | `CountsSkippedBlankLinesTowardTheEstimate` |
| B — `estimatedBytes += line.Length` (drop the terminator) | the 3 forward-only counter cases |
| C — always report the estimate, ignore `CanSeek` | `ReportsTheStreamsRealPositionNotTheEstimate` |
| D — always report `stream.Position`, ignore `CanSeek` | the 3 forward-only counter cases |
| E — stop skipping blank lines | `DropsBlankAndWhitespaceOnlyLines`, `OnAStreamOfNothingButBlankLines`, `CountsSkippedBlankLines…` |
| F — `leaveOpen: false` | `LeavesTheCallersStreamOpen` + both lease cases |
| G — seek to 0 before reading | `StartsAtTheStreamsCurrentPositionNotAtZero` |
| H — drop `ct` from `ReadLineAsync` | `WithAnAlreadyCancelledToken_Throws` |
| I — leak the lease instead of `await using` | both `ReleasesTheLease…` cases |
| J — `ToAsyncEnumerable` walks its source eagerly | `DoesNotTouchTheSourceUntilItIsEnumerated` |
| K — `ToAsyncEnumerable` reverses its source | `PreservesOrderAndContent` |

## What a reviewer might push back on

Two cases were **written and then deliberately removed** rather than shipped, because neither could be made to fail:

- an empty-stream case (`Assert.Empty`) survived all eleven mutations — there is no realistic way to break the reader that makes a zero-byte stream yield a line, so it was carrying no information. The blank-lines-only case covers the same "yields nothing" shape and *is* killed by mutation E.
- a "rewinds to the source's origin, not byte zero" case, which turned out to be asserting `SdCardParseSource.Open()`'s contract rather than this class's — already pinned by `SdCardParseSourceTests` from #654.

The seekable-stream case does depend on the whole payload fitting in one `StreamReader` buffer. That is stated in a comment on the test; the assertion is on the reader's contract (report the stream's real position), and the buffering only determines what that position happens to be.

Verification: full solution suite green under `-warnaserror` on net9.0 and net10.0 — 3915 passed / 2 skipped (hardware) on each, plus 217 in `Daqifi.Mcp.Tests`. Zero warnings. Bench validation does not apply and was not skipped silently: this PR changes no production code and performs no device interaction.

**Not merging — for review.**

### DIFF
--- a/src/Daqifi.Core.Tests/Device/SdCard/SdCardTextLineReaderTests.cs
+++ b/src/Daqifi.Core.Tests/Device/SdCard/SdCardTextLineReaderTests.cs
@@ -1,0 +1,299 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Daqifi.Core.Device.SdCard;
+using Xunit;
+
+namespace Daqifi.Core.Tests.Device.SdCard;
+
+/// <summary>
+/// Direct tests for <see cref="SdCardTextLineReader"/>, the line pump behind the CSV and JSON
+/// log parsers.
+/// </summary>
+/// <remarks>
+/// <para>
+/// Every non-blank line of a downloaded text log reaches a parser through this class, carrying
+/// the running byte count that drives the caller's progress reporting. It had no tests of its
+/// own: the CSV and JSON parser fixtures feed it small, well-formed, seekable
+/// <see cref="MemoryStream"/>s positioned at zero and assert on the samples that come out the
+/// far end, so they never observe the byte counter at all and never take the non-seekable
+/// branch.
+/// </para>
+/// <para>
+/// That left the parts that decide what a progress bar shows unpinned — that a seekable source
+/// reports the stream's real position while a forward-only one reports an estimate, that blank
+/// lines are dropped from the output but still counted toward that estimate, and that reading
+/// starts where the caller left the stream rather than at byte zero. It also left the lease
+/// discipline of the source overload unpinned: that overload is the only thing that returns a
+/// stream-backed source's single read cursor, and it has to return it when the consumer stops
+/// early as well as when the file runs out.
+/// </para>
+/// </remarks>
+public class SdCardTextLineReaderTests
+{
+    private const string Content = "alpha\nbravo\ncharlie\n";
+
+    #region Blank-line handling and the byte counter
+
+    [Fact]
+    public async Task ReadLines_DropsBlankAndWhitespaceOnlyLines()
+    {
+        using var stream = new MemoryStream(Encoding.ASCII.GetBytes("alpha\n\nbravo\n   \n\t\ncharlie\n"));
+
+        var lines = await ReadAllAsync(stream);
+
+        Assert.Equal(new[] { "alpha", "bravo", "charlie" }, lines.Select(l => l.Text));
+    }
+
+    [Fact]
+    public async Task ReadLines_OnAForwardOnlyStream_CountsSkippedBlankLinesTowardTheEstimate()
+    {
+        // "alpha\n" is 6 bytes and the blank line 1 more, so "bravo" must be reported at 13 —
+        // the 12 it would carry if the blank were skipped before the counter ran is wrong.
+        using var stream = new ForwardOnlyStream(Encoding.ASCII.GetBytes("alpha\n\nbravo\n"));
+
+        var lines = await ReadAllAsync(stream);
+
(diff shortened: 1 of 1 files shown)


tylerkron:

Independently re-verified on head `11ebc52` (no production diff; the only changed file is the new test):

- **Mutation evidence re-run from a pristine copy, not taken on trust.** Six independent mutant builds — C, A, I, G, E+B, and F+H+J+K — killed 1, 1, 2, 1, 5 and 6 cases respectively. The three spot-checked singles matched the A–K table exactly, and the union across all six runs is **all 13 cases**, so no test in the file is unfailable. The pristine file was restored byte-identical (`git status` clean).
- **No weak shapes.** No constant assertions, mock-return round-trips, or auto-property round-trips; no assertion made vacuous by an earlier one in the same test.
- **No overlap with #654.** The two lease cases pin *this* class's `await using` on the lease (mutation I kills both), whereas `SdCardParseSourceTests` pins the source's own latch (`Open_While…Outstanding_Throws` / `Open_AfterTheOutstandingLeaseIsDisposed_Succeeds`). `StartsAtTheStreamsCurrentPositionNotAtZero` runs over a raw `MemoryStream`, not a source, so it is not a restatement of `Open_RewindsToWhereTheStreamWasWhenItWasWrapped_NotToByteZero`.
- **Re-validated against current `main`** (`c546abd`, after #650/#652/#654/#656/#657). Test-merge is a no-op — the branch already contains `origin/main` — and the full solution suite is green under `-warnaserror`: net9.0 3915 passed / 2 skipped, net10.0 3915 passed / 2 skipped, `Daqifi.Mcp.Tests` 217 passed. Zero warnings.
- **Qodo**: 1 round, `Bugs (0) / Rule violations (0) / Requirement gaps (0)`, and **both** surfaces checked — the summary comment *and* an unfiltered `reviewThreads` dump returning `totalCount: 0`. Settled: re-checked against the same head SHA well over 3 minutes later with no change to the review body's timestamp and still zero threads. The zero-thread result was control-validated against **#659**, which returns 1 thread through the same query, so it is not an empty-query artefact.
- **No collision** with #668 (the SD-card `ISampleSource` work) — disjoint file sets.

Separately, the `HidStringNormalization.Normalize("\0") == ""` wart noted in the handover is still live on `main` and has been filed on its own as #670. It is deliberately **not** folded into this PR.

PR #671, opened 2026-08-24
test(export): census the range guards in Logging/Export so they cannot drift apart

### LINKED ISSUE #664: Guard-uniformity census stops at the SCPI producers: range guards elsewhere (incl. #657's new ones) are uncensused
## Summary

PR #660 (`closes #651`) adds `EveryValueRangedProducer_RejectsOutOfRangeArgumentIdentically` — a structural census that walks **all nine** inline range-guard sites in `ScpiMessageProducer` and asserts each throws `ArgumentOutOfRangeException` with a matching `ParamName`, a populated `ActualValue`, and the expected message prefix. It is a good test: it locks the shape of a guard family so no member can drift.

The census is deliberately scoped to the SCPI producers. That scope is correct for #660 and should not change as part of it. But it means the uniformity guarantee stops at one file, while range guards of exactly the same shape are spread across the rest of `Daqifi.Core` — and new ones keep arriving.

**This is not a defect in #657 or #660. Neither needs any change to merge.** This issue only records a gap that becomes visible once both land.

## What triggered this

PR #657 (`closes #639`, live-stream-to-CSV recording) adds two new range guards in `RecordLiveSamplesToCsvAsync`, `src/Daqifi.Core/Logging/Export/LiveCsvRecording.cs`:

```csharp
if (duration is { } window && window <= TimeSpan.Zero)
{
    throw new ArgumentOutOfRangeException(
        nameof(duration), window, "Recording duration must be greater than zero.");
}

if (bufferCapacity is { } capacity && capacity < 1)
{
    throw new ArgumentOutOfRangeException(
        nameof(bufferCapacity), capacity, "Buffer capacity must be at least 1.");
}
```

I checked these against each other and against the repo's established conventions — the shared `ValidateChannel` / `RequireFinite` helpers and the `UseAdcCalibration` (`bank is < 0 or > 1` → `"Calibration bank must be 0 (factory) or 1 (user)."`) precedent:

- **They are mutually consistent** and **conventional**: both use the three-argument constructor, so `ParamName` matches the caller's own parameter and `ActualValue` is populated; both messages are sentence-cased, name the subject, and end with a period. Each has a dedicated `[Theory]` test in `LiveCsvRecordingTests` asserting the exception type and `ParamName`. If the census were widened tomorrow, these two would pass it as written.

So the guards are fine. The point is that nothing would have caught it if they weren't — nothing outside `ScpiMessageProducerTests` asserts guard uniformity at all.

## Real scope of the gap

A quick `rg` for `ArgumentOutOfRangeException(` over `src/Daqifi.Core`, excluding `ScpiMessageProducer.cs`, finds **82 throw sites** today (#657 makes 84). They span `Device/`, `Device/SdCard/`, `Device/Discovery/`, `Device/Internal/`, `Channel/`, `Communication/Transport/`, `Firmware/`, `Firmware/Winc/`, and `Logging/Export/`. The census covers 9 of them.

That population is already not uniform. Two concrete divergences visible from the sweep:

1. **Dropped `ActualValue`.** The census asserts `ex.ActualValue` on every SCPI site, but a number of non-SCPI guards use the two-argument `(paramName, message)` constructor and therefore report `ActualValue == null` — e.g. `Channel/DigitalChannel.cs:196`, `Channel/AnalogChannel.cs:280`, `Communication/Transport/HidLibraryTransport.cs:265`, `Device/TimestampProcessor.cs:95`, `Device/SdCard/SdCardFileParser.cs:54`, `Device/SdCard/SdCardFileReceiver.cs:118`.
2. **Composed `ParamName`.** The pre-existing guard in `Logging/Export/CsvExporter.cs:36` throws with `ParamName` `"options.AverageWindow"` rather than a plain `nameof`. Arguably right for a nested option, but it is a second convention living in the same folder as the two new #657 guards, and nothing states which one a future author should copy.

Neither of these is urgent or a bug. They are exactly the kind of quiet drift the #651 census exists to prevent — happening in the places the census does not look.

## Why it matters

The census's value is that adding a tenth SCPI guard in a non-conforming shape turns it red. Add a guard anywhere else in the codebase and nothing turns red. Since new guards are landing outside the SCPI producers (#657 is the proof), the guaranteed-uniform region is not where the growth is.

## Suggestion

Add a **companion census** rather than widening the existing one. Widening `EveryValueRangedProducer_RejectsOutOfRangeArgumentIdentically` would break its clean 1:1 mapping onto one file, and the guards elsewhere are not all statically invokable the way the `ScpiMessageProducer` statics are.

Concretely, either or both of:

- **A narrow export/logging census** over `Logging/Export/` — the two #657 guards plus `CsvExporter`'s `AverageWindow` guard — in the same table-driven style as #660: name, `ParamName`, `ActualValue`, message prefix, and an `Assert.Equal(n, entries.Length)` count so a new guard added without a census entry fails the count. Small, cheap, mirrors #660, and settles the `options.AverageWindow` vs `nameof` question for that folder.
- **A reflection-based sweep** asserting the invariant that does hold repo-wide and needs no per-site table: every `ArgumentOutOfRangeException` thrown from public API surface carries a non-null `ParamName` matching a real parameter, a non-null `ActualValue`, and a message ending in `.`. Higher coverage, no maintenance burden as guards are added, but only if it can be driven without an unreasonable amount of scaffolding to reach each site.

The first is the safe, proportionate step. The second is worth a look only if someone wants the whole 84-site population under a guarantee.

## References

- #651 — the census request
- #660 — adds the nine-site SCPI census (open, ready for review; **no change needed**)
- #657 — adds the two `LiveCsvRecording` guards (open, ready for review; **no change needed**)

tylerkron:
### What was wrong

When Core rejects an out-of-range argument, what the caller actually gets back is inconsistent. Some guards throw the two-argument `ArgumentOutOfRangeException`, which leaves `ActualValue` null — so the value that was rejected simply isn't reported — while others use the three-argument form and do report it. Nobody notices, because nothing checks. `ScpiMessageProducer` got a census in #660 that locks its nine guards into one shape, but that's one file, and the new guards aren't landing there: #657 added two in `Logging/Export`, and #668 added another while this branch was open.

So the region that's guaranteed uniform is not the region that's growing, and a caller writing `catch (ArgumentOutOfRangeException ex)` can't rely on `ex.ActualValue` being populated.

### How it was fixed

A companion census over `Logging/Export` — the folder where the recent guards actually landed, four sites today. It invokes each guard and asserts the exact `ParamName`, `ActualValue` and message the caller sees, and that the message is a sentence ending in a period. Then it reads the folder's source and asserts the table matches every `ArgumentOutOfRangeException` site it finds, per file, so a guard added without a census entry fails loudly instead of being quietly uncovered.

**The thing a reviewer is most likely to want to argue with:** `CsvExporter` composes its `ParamName` as `"options.AverageWindow"` where everything else uses a plain `nameof`. That's the one real divergence in scope, and I pinned it as-is rather than normalising it, with a comment saying why — `ParamName` is observable, a caller can filter on it, and the composed form says *which* option was rejected where `nameof(options)` wouldn't. To stop that becoming a loophole, the census resolves every `ParamName` structurally: the first segment must name a real parameter of the throwing member and the tail a real public member of that parameter's type, so a rename that leaves the string behind fails.

The divergent guards *outside* this folder (`DigitalChannel`, `AnalogChannel`, `HidLibraryTransport`, `TimestampProcessor`, the two SD-card files) are out of scope and stay recorded in #664.

### Verification

The census can fail, and I checked, by mutating production code and reverting:

- Two-argument constructor on `LiveCsvRecording`'s `bufferCapacity` guard — i.e. drifting it into the shape six other guards in the repo already have — turns the census **red on `ActualValue`**, while every other test in the export test classes stays green, including `LiveCsvRecordingTests`' own `bufferCapacity` test. That's the point: the existing per-site tests assert type and `ParamName` only, so this drift is invisible to them.
- An extra uncensused guard in `CsvExporter` turns the source scan **red**, naming the file.
- It also catches a whole **new file**, not just new guards in existing ones — which is not hypothetical: #668 landed `SdCardLogSampleSource` with a constructor guard while this branch was open, the scan went red naming it, and `59211a6` adds the row (widening `GuardSite` from `MethodInfo` to `MethodBase` so a constructor censuses on the same terms).

Full suite green on net9.0 and net10.0 under `-warnaserror`: 3982 passing, 0 failing, plus `Daqifi.Mcp.Tests`. Bench validation doesn't apply — test-only change, no production code touched.

`closes #664`

Not merging — for review.


tylerkron:
Both findings taken, fixed in e4ab807.

**1. Guard scan matches non-throws — agreed, and it was the more serious of the two.** The scan counted any line mentioning `ArgumentOutOfRangeException` outside a `//` or `*` line, so a `catch`, a `typeof`, or a mention inside a `/* */` block would have failed the census for a change that added no guard. A drift test that cries wolf is one people switch off. It now matches `throw new ArgumentOutOfRangeException` and the `ArgumentOutOfRangeException.ThrowIf*` helpers only (keeping helper-form coverage, which was the reason the match was loose), over comment-stripped lines that carry block-comment state across the file. Re-ran the mutation that adds an uncensused guard to `CsvExporter` against the tightened scan: still red, same message.

**2. Reflection lookup brittle — agreed.** `Type.GetMethod(name)` would throw `AmbiguousMatchException` on the first overload added, which tells the reader nothing. Resolution is now by enumeration with an assertion whose message says what to do: the census must name which overload its guard lives in.

Full suite green on net9.0 and net10.0 under `-warnaserror` after the change (3905 passing, plus Daqifi.Mcp.Tests).

tylerkron:

Independently re-verified at head `e4ab807` (already contains current main `c546abd`, so the test-merge is a fast-forward):

- **Both claimed mutations reproduce.** Two-arg constructor on the `bufferCapacity` guard → census red on `ActualValue` (`Expected (…, "bufferCapacity", 0)` / `Actual (…, "bufferCapacity", null)`), with all 75 other tests in the three export test classes green, including `LiveCsvRecordingTests`' own `bufferCapacity` test. An extra uncensused guard in `CsvExporter` → source scan red (`CsvExporter.cs: 1` vs `2`).
- **Round 1's scan fix is correctly calibrated — not overcorrected.** A file containing only non-guard mentions (`<exception cref>` doc tag, `catch`, `typeof`, `is`, and a block comment literally spelling out `throw new ArgumentOutOfRangeException`) yields zero sites: no false positives. The one blind spot is a `throw new` dangling on its own line before the type name; zero of the repo's throw sites are written that way, so it is theoretical rather than live. The scan is also non-recursive, which matches a folder that has no subdirectories.
- **The scan notices a new file, not just new guards.** Dropping a new `.cs` file with a conventional guard into the folder turns the scan red and names the file.
- **Divergence documentation checked.** The `"options.AverageWindow"` note states why it is pinned (ParamName is observable contract; the composed path says *which* option was rejected) and the structural resolve keeps it honest. It reads as a decision, not an oversight.
- Full suite under `-warnaserror`: net9.0 3905 / net10.0 3905 / `Daqifi.Mcp.Tests` 217, 0 failures. CI `build` success on `e4ab807`.
- Qodo round 2 confirmed on **both** surfaces at this head: `Code Review by Qodo` at Bugs (0) / rule violations (0) / skill insights (0) with both round-1 findings struck through, and an unfiltered `reviewThreads` dump returning 2 threads, both `isResolved: true` — a non-empty result, so the query is proven live rather than silently filtering. Settled 15 minutes; unchanged.

**One thing reviewers should weigh before merging: the merge order with #668 has a real cost.** Confirmed against #668 at head `2c1de9b` — it adds `SdCardLogSampleSource.cs` to this very folder with an `analogChannelCount` guard. The census **will go red for whichever of the two merges second**; it does not silently ignore the new guard. The row cannot be pre-added here (`typeof(SdCardLogSampleSource)` does not compile until #668 lands), and loosening the scan to tolerate a known-pending file is exactly what would make the census vacuous, so it is deliberately not papered over. Recommended order and the exact two-part follow-up — the new guard sits in a **constructor**, so `GuardSite.Method` needs widening from `MethodInfo` to `MethodBase` as well as the new row — are now written into the PR body.

Not merging.

PR #691, opened 2026-08-27
docs(capabilities): warn that the advertised PWM minimum is not usable

### LINKED ISSUE #684: Capability document advertises a 1 Hz PWM floor that Core itself rejects (and the hardware cannot produce)
## What a consumer hits

`CapabilityChannel.PwmMinimumFrequencyHz` is documented as *"the lowest PWM frequency the pin
accepts"*. On every shipping Nyquist the device advertises **1 Hz** there — so the obvious thing to
write is:

```csharp
var pin = doc.Channels.First(c => c.SupportsPwm);
device.SetPwmFrequency(pin.PwmMinimumFrequencyHz!.Value);   // 1
```

…and Core throws:

```
ArgumentOutOfRangeException: PWM frequency must be 6-50000 Hz. (Parameter 'frequencyHz')
Actual value was 1.
```

So the library advertises a floor it then refuses. And this is not just Core being conservative:
Core's own comment on `DaqifiStreamingDevice.MinPwmFrequencyHz` says that below 6 Hz *"the
firmware's 16-bit period register silently wraps and the output runs in the kilohertz range"* — a
consumer who bypassed Core and sent 1 Hz over SCPI would get a kilohertz square wave, not a 1 Hz
one. The advertised value is wrong in the firmware, and Core repeats it to consumers as if it were
usable.

## Bench evidence

Nq1, fw 3.7.2, sn `9090539562006014104`, `/dev/cu.usbmodem1101`, via `CONFigure:CAPabilities:JSON?`:

```
pwm-capable pins: [0,3,4,5,6,7]  advertised min=1  max=50000
Core MinPwmFrequencyHz=6  MaxPwmFrequencyHz=50000
SetPwmFrequency(1) -> ArgumentOutOfRangeException
```

The PWM-capable pin set `{0,3,4,5,6,7}` and the 50 kHz ceiling both match Core exactly; only the
floor disagrees.

## Suggested fix

Documentation only — no API change, no behavior change. Neither `PwmMinimumFrequencyHz` nor
`PwmMaximumFrequencyHz` is read anywhere inside Core (`rg` confirms: parser writes them, nothing
else reads them), so they are purely informational surface for consumers, and the parse itself is
correct — the document really does say 1.

On `CapabilityChannel.PwmMinimumFrequencyHz`, say that the device under-reports this: it advertises
1 Hz on all supported firmware, but frequencies below `DaqifiStreamingDevice.MinPwmFrequencyHz` (6)
are not reproduced correctly by the hardware, and `SetPwmFrequency` rejects them. Point the reader
at `MinPwmFrequencyHz` as the value to actually use as a floor. Worth a matching note wherever the
capability document is described as the device's self-description, since this is the one field on it
that is known-wrong rather than merely absent.

Found during a non-destructive bench validation pass over the capability-document surface. The rest
of that surface checked out: the parse round-trips every field against an independent
`System.Text.Json` re-read, the merge overlays board-derived capabilities correctly (ai=16, ao=0,
dio=16, maxRate 1000 -> 22000), and `SampleRateCap` tracked the device's own `current_max_rate_hz`
exactly across four different enabled-channel sets.

tylerkron:
## What was wrong

If you asked a Nyquist what its lowest PWM frequency is, it said 1 Hz — and Core repeated that
answer to you through `CapabilityChannel.PwmMinimumFrequencyHz`, documented as "the lowest PWM
frequency the pin accepts". Write the obvious thing:

```csharp
device.SetPwmFrequency(pin.PwmMinimumFrequencyHz!.Value);   // 1
```

…and Core throws `ArgumentOutOfRangeException: PWM frequency must be 6-50000 Hz`. The library was
advertising a floor it then refused. Worse, the firmware is the one that's wrong here: below 6 Hz
the 16-bit period register wraps and the pin actually outputs a kilohertz square wave, so a
consumer who bypassed Core and sent 1 Hz over raw SCPI would get silently bad output rather than
an error.

## How it was fixed

Documentation only. `PwmMinimumFrequencyHz` now says plainly not to use it as a floor, explains
why the device under-reports it, and points at `DaqifiStreamingDevice.MinPwmFrequencyHz` (6) as
the value that actually works. `CapabilityDocument`'s own remarks — which describe it as the
device's accurate self-description — now carry a matching note, since this is the single field on
that document known to be wrong rather than merely absent.

No API or behavior change was made. The parse is correct (the document really does say 1), and
nothing inside Core reads either PWM frequency field, so they are purely informational surface for
consumers. Fixing the value rather than the docs would mean Core lying about what the device said,
which is worse; the honest move is to keep the reported value and label it.

## Verification

Full suite green on net9.0 and net10.0 (4066 Core + 217 MCP tests each), 0 build warnings.
Bench evidence for the discrepancy is in the issue: Nq1, fw 3.7.2, `CONFigure:CAPabilities:JSON?`
reports min=1 / max=50000 with PWM pins `{0,3,4,5,6,7}` — only the floor disagrees with Core.

closes #684

### DIFF
--- a/src/Daqifi.Core/Device/Capabilities/CapabilityChannel.cs
+++ b/src/Daqifi.Core/Device/Capabilities/CapabilityChannel.cs
@@ -58,7 +58,19 @@ public sealed class CapabilityChannel
     /// </summary>
     public bool SupportsPwm { get; init; }
 
-    /// <summary>Gets the lowest PWM frequency the pin accepts, or <c>null</c> when not stated.</summary>
+    /// <summary>
+    /// Gets the lowest PWM frequency the pin advertises, or <c>null</c> when not stated.
+    /// </summary>
+    /// <remarks>
+    /// Do not use this as a floor. The device under-reports it: every supported firmware
+    /// advertises 1 Hz here, but below
+    /// <see cref="DaqifiStreamingDevice.MinPwmFrequencyHz"/> (6 Hz) the firmware's 16-bit
+    /// period register silently wraps and the output runs in the kilohertz range, so a commanded
+    /// 1 Hz is not what the pin produces. <see cref="DaqifiStreamingDevice.SetPwmFrequency"/>
+    /// rejects anything below that constant, which is the value to use as the real floor. This is
+    /// the one field on the capability document known to be wrong rather than merely absent —
+    /// <see cref="PwmMaximumFrequencyHz"/> and the set of PWM-capable pins both match the hardware.
+    /// </remarks>
     public int? PwmMinimumFrequencyHz { get; init; }
 
     /// <summary>Gets the highest PWM frequency the pin accepts, or <c>null</c> when not stated.</summary>
--- a/src/Daqifi.Core/Device/Capabilities/CapabilityDocument.cs
+++ b/src/Daqifi.Core/Device/Capabilities/CapabilityDocument.cs
@@ -21,6 +21,13 @@ namespace Daqifi.Core.Device.Capabilities;
 /// property that is <c>null</c> means "this document did not state it", never "the device does not
 /// have it".
 /// </para>
+/// <para>
+/// The document is the device's own account of itself, and it is accurate everywhere Core has
+/// checked it against hardware — with one exception. <see cref="CapabilityChannel.PwmMinimumFrequencyHz"/>
+/// advertises 1 Hz on every supported firmware, a frequency the hardware cannot actually produce;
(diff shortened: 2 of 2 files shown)


tylerkron:

Head `98308ab`: 1 Qodo round, zero findings ("no issues found"), 0 unresolved review threads on that head, all four CI jobs green. Re-checked both surfaces well after the round settled. Documentation-only change (XML doc comments on `CapabilityChannel`/`CapabilityDocument`), so no bench validation applies — no runtime behavior changed.

Issue #637, opened 2026-08-24
test: adopt TimeProvider — seven flake tickets and 170 real sleeps all trace to one missing clock seam


tylerkron:
**Type**: Architecture / Test reliability
**Priority**: High

## What
Every timeout, deadline, and retry backoff in `Daqifi.Core` reads the ambient system clock directly. There is no injectable clock anywhere in the repo:

```
$ grep -rn "TimeProvider" src --include="*.cs" --include="*.csproj" | wc -l
0
```

What is there instead:

- **41** direct `DateTime.UtcNow` / `DateTime.Now` reads in `src/Daqifi.Core`
- **73** hardcoded `TimeSpan.FromSeconds` / `FromMilliseconds` constants
- **53** `Task.Delay` calls and **9** `Thread.Sleep` calls
- `Stopwatch` in `TextExchangeEngine`, `SdCardOperations`, `DaqifiDevice`, `DeviceErrorThrottle`, `ProcessExternalProcessRunner`

The canonical shape, `Device/Internal/TextExchangeEngine.cs:514-533` — a wall-clock deadline advanced by a 50 ms real sleep:

```csharp
var lastMessageTime = DateTime.UtcNow;
var startTime = DateTime.UtcNow;
while (DateTime.UtcNow - startTime < maxWait)
{
    await Task.Delay(50, cancellationToken).ConfigureAwait(false);
    ...
    var elapsed = DateTime.UtcNow - lastMessageTime;
```

Because the production code can only be driven by real time, the tests can only assert against real time. `src/Daqifi.Core.Tests` contains **170 `Task.Delay`/`Thread.Sleep` calls across 45 files**, concentrated exactly where the flakes are:

```
SdCardOperationsTests.cs                    15
DeviceReconnectTests.cs                     10
DeviceErrorSurfaceTests.cs                  10
DaqifiDeviceRawCaptureLockTests.cs          10
DaqifiDeviceOperationSerializationTests.cs  10
```

## Why
This is the single root cause behind a recurring, self-refilling ticket class. Every one of these is a wall-clock race, and they keep coming back:

| Issue | State |
|---|---|
| #634 — `..._DoesNotWaitOutTheTimeout` has an over-tight 2000 ms bound | **open** |
| #632 — `..._KeepsABlankThatArrivesOnceTheCommandIsBeingWritten` races on CI | closed |
| #589 — stop asserting an exact retry count in a wall-clock race | closed |
[...]

PR #696, opened 2026-08-28
test(sd): stop the abandoned-download test racing a production delay against its own deadline

### LINKED ISSUE #695: CI: the abandoned-SD-download test races a 100ms production delay against its own 300ms deadline, and lost on main
Seen on [run 33109464109](https://github.com/daqifi/daqifi-core/actions/runs/33109464109) — a `push` run on **main** at `61348f8`, so this is a red main rather than a red PR. `test (macos-latest)` failed; the other two legs and both frameworks were fine.

```
Daqifi.Core.Tests.Device.SdCard.SdCardOperationsCollaboratorTests
  .DownloadSdCardFileAsync_TransferThatIgnoresItsToken_IsAbandonedWithoutARestore [FAIL]
  Failed  ... [944 ms]
  Error Message:
   Assert.Contains() Failure: Item not found in collection
```

Nothing in `61348f8` (`Device/Diagnostics/*`) touches the SD path, so this is a flake, not a regression.

## Mechanism

The failing assertion is the one that says the transfer got as far as issuing the GET:

```csharp
Assert.Contains("send:SYSTem:STORage:SD:GET \"data.bin\"", host.Calls);
```

The test drives a transfer that ignores its cancellation token and asserts it is abandoned by the download's hard deadline. It configures a **200 ms** budget, which `HardDeadlineFor` turns into a **300 ms** hard deadline.

But before the GET goes out, `DownloadSdCardFileAsync` awaits a fixed **100 ms** SD-interface settle delay:

```csharp
PrepareSdInterface();
await Task.Delay(SD_INTERFACE_SETTLE_DELAY_MS, ct).ConfigureAwait(false);   // 100 ms
_host.Send(ScpiMessageProducer.GetSdFile(fileName));
```

So the test leaves **200 ms of margin** between "the settle delay finishes" and "the deadline fires". That delay's continuation resumes on the shared thread pool — the same pool the rest of the suite is competing for — while the deadline runs on its own timer and does not slip with it. On a loaded macOS runner the continuation lands late, the deadline wins, the transfer is abandoned before the GET is ever sent, and the assertion fails.

The failure signature matches exactly: it is the only assertion in the test that requires the transfer to have made progress past that delay, and the test's own three "nothing was restored" assertions are all `DoesNotContain`, which pass vacuously in that situation.

The sibling parked-transfer tests are not exposed to this. `DownloadSdCardFileAsync_WhileAnotherIsParkedOnTheTransport_RefusesInsteadOfSharingIt` anchors on the parked read rather than a budget, and the `ParkedDownloadDevice` family in `SdCardOperationsTests` only asserts that a `TimeoutException` came back, which holds whether or not the GET went out.

## Suggested fix

Two changes, both in the test:

1. Give the budget real headroom over the 100 ms delay it has to cover — 2 s, i.e. a 2.2 s deadline. It is headroom, not pacing; nothing waits on it deliberately.
2. Anchor the assertions on the transfer having reached the parked read (the existing, currently-unused `entered` TCS) instead of on the clock. That read is entered only after the GET is sent, so it is a direct statement of the precondition, and a future timing failure then reports itself as a timing failure rather than as a missing command.

## Related

- #687, #688 — the same class of macOS-CI timing flake, and the same fix shape (anchor on the event, keep timeouts as deadlock guards).
- #689 — the other open intermittent CI failure, unrelated cause (coverlet's post-test step).

tylerkron:
## What was wrong

`main` went red on a run where **every test passed**. The macOS leg of the CI run for `61348f8` failed on one SD-card test, `DownloadSdCardFileAsync_TransferThatIgnoresItsToken_IsAbandonedWithoutARestore`, with `Assert.Contains() Failure: Item not found in collection`. Nothing in that commit goes anywhere near the SD path — it was a flake, and the same class of flake has already dequeued two PRs from the merge queue this week.

The test itself was the problem. It gives the download a 200 ms budget (a 300 ms hard deadline) and then asserts the transfer got far enough to send its `SD:GET`. But the download waits a fixed 100 ms for the SD interface to settle before sending anything, and that wait resumes on the thread pool the rest of the suite is competing for, while the deadline ticks on its own timer. On a loaded runner the deadline wins, the transfer is abandoned before the command ever goes out, and the assertion fails — reporting a missing command when nothing is missing.

## How it was fixed

Test-only. The budget becomes 2 s — twenty times the delay it has to outlast, as headroom rather than as pacing — and the assertions are now anchored on the transfer having actually reached its parked read instead of on the clock. That read is entered only after the `SD:GET` is sent, so waiting for it states the precondition directly, and a future timing failure reports itself as a timing failure rather than as a missing command. The signal was already there in the test; it was created and never awaited.

The one thing worth pushing back on: the test now takes about 2.2 s instead of about 0.3 s, because it has to sit through the longer budget for the deadline to fire. That is the cost of the margin. The alternative — keeping it short and hoping the runner is quick — is what just turned main red.

## Verification

Full suite green on net9.0 and net10.0 (4092 + 217 tests per framework). The fixed test also passes under deliberate CPU starvation locally; the original flake did not reproduce on this machine, which has four times the runner's cores, so the mechanism is established by reading the timing rather than by a local repro — the issue lays it out in full.

No production code changed, so no bench validation applies.

closes #695

Not merging — this is for your review.

### DIFF
--- a/src/Daqifi.Core.Tests/Device/SdCard/SdCardOperationsCollaboratorTests.cs
+++ b/src/Daqifi.Core.Tests/Device/SdCard/SdCardOperationsCollaboratorTests.cs
@@ -1183,8 +1183,21 @@ public class SdCardOperationsCollaboratorTests
         var host = new FakeHost
         {
             IsStreaming = true,
-            // 200ms budget → a ~300ms hard deadline (the grace is clamped to a 100ms floor).
-            SdCardDownloadTimeout = TimeSpan.FromMilliseconds(200),
+            // 2s budget → a 2.2s hard deadline (the grace is 10% of the budget, floored at 100ms).
+            //
+            // The deadline has to outlast the transfer's own run-up to the GET, which is a fixed
+            // 100ms SD-interface settle delay the download awaits before sending anything. That
+            // delay resumes on the shared thread pool, so on a loaded runner it can finish far
+            // later than 100ms after it was started, while the deadline runs on its own timer and
+            // does not slip with it. The previous 200ms budget left only 200ms of margin for that
+            // slip, and lost: the deadline fired first, the transfer was abandoned before the GET
+            // ever went out, and the assertion below failed on macOS CI with "Item not found in
+            // collection" — after every test in the run had passed, so it turned main red rather
+            // than reporting a real defect.
+            //
+            // 2s is not a pacing knob and nothing here waits on it deliberately: it is headroom,
+            // twenty times the delay it has to cover.
+            SdCardDownloadTimeout = TimeSpan.FromSeconds(2),
             RawCaptureStreamFactory = () => new ParkedStream(entered, release.Task),
         };
         var ops = new SdCardOperations(host);
@@ -1194,7 +1207,7 @@ public class SdCardOperationsCollaboratorTests
         // MemoryStream needs no disposal, and disposing it here would only manufacture a race.
         var destination = new MemoryStream();
 
-        // Bounded well above the ~300ms hard deadline this test asserts. The stream deliberately
+        // Bounded well above the 2.2s hard deadline this test asserts. The stream deliberately
         // ignores its token, so if the deadline itself regressed the download would never return,
         // and an unbounded await would hang the whole run instead of failing the one test that
         // caught it. The bound is a token rather than WaitAsync(TimeSpan) on purpose: the latter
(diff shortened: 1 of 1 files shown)


tylerkron:

Head `5b46c83`: 1 Qodo round, zero findings ("no issues found", Bugs 0 / Rule violations 0 / Requirement gaps 0), 0 unresolved review threads. Re-checked both surfaces after the round settled; head unchanged. All four CI jobs green, including the macOS leg this test failed on.

Qodo's one substantive note was an alternative approach rather than a finding — inject a clock seam so the settle delay and the deadline become deterministic and the test stays fast — and it recommended against it here, since that means a production-facing abstraction for one regression test. Worth knowing that #637 (adopt `TimeProvider`) is exactly that seam; if it lands, this test's 2 s budget is one of the things it can take back.

Test-only change, so no bench validation applies.

Issue #697, opened 2026-08-28
perf(sdcard): the protobuf log parser spends 1.6ms and 6.7MB before its first sample, while CSV and JSON take 2us


tylerkron:
## What a consumer hits

Ask the three SD-card log parsers for the *first* sample of a 10,000-row recording and two of them
answer in about two microseconds. The protobuf `.bin` parser takes **1.6 ms and allocates 6.7 MB**
before it yields anything — a third of the time, and 40% of the allocation, of draining the whole
file.

Measured on the new benchmark harness (#640), 8 analog channels, ~880 KB file:

| Method | Mean | Allocated |
| --- | ---: | ---: |
| ProtobufDrainAll | 4,800 µs | 16,880 KB |
| **ProtobufTimeToFirstSample** | **1,627 µs** | **6,754 KB** |
| CsvDrainAll | 5,556 µs | 15,212 KB |
| CsvTimeToFirstSample | 1.96 µs | 13.6 KB |
| JsonDrainAll | 7,040 µs | 7,181 KB |
| JsonTimeToFirstSample | 2.09 µs | 9.9 KB |

Issue #489 made these parsers lazy so a caller could start writing rows immediately instead of
waiting for the whole file. For CSV and JSON that landed. For protobuf it did not, and nothing
noticed because "lazy" was measured as "does not materialize the whole file" rather than as a
number.

## Why

Two mechanisms, and neither is the one you would guess.

**It is not the configuration pre-scan's message limit.** `SdCardFileParser.BuildSessionAsync`
reads a prefix of the log to find the device configuration, bounded by
`SdCardParseOptions.ConfigurationScanMessageLimit` (512 by default). Dropping that to 8 changed
the measurement by nothing at all: same 1.6 ms, same 6.6 MB.

**It is `SdCardParseOptions.BufferSize`.** `ReadMessagesAsync` reads `BufferSize` bytes (64 KB by
default), hands the whole chunk to `ProtobufMessageParser.ParseMessages`, and only then yields the
messages one at a time. So the first sample costs a full parse of ~1,300 messages regardless of
the scan limit — the limit is applied *after* the chunk has already been parsed. And because the
configuration pass and the sample pass each open the log from the start, that first 64 KB is
parsed twice.

That also explains the ratio: one chunk is ~13% of this file, parsed twice is ~26%, which is
roughly the 34% measured.

## Suggested fix

Nothing about the file format forces this — the parser could yield each message as it is decoded
[...]

Issue #491, opened 2026-08-12
perf(device): idle connected devices burn CPU — producer poll wakeups, per-device port enumeration every second, a new OS thread per SCPI exchange


tylerkron:
**Type**: Performance / Resource
**Priority**: Medium

## What
Idle connected devices burn CPU and threads:

1. **Producer poll** — `MessageProducer.cs:188-189`: `_messageAvailable.Wait(100); _messageAvailable.Reset();` wakes 10×/s forever per device. Both `Stop()` and `StopSafely()` already `Set()` the event before joining — the 100 ms timeout exists only to paper over the lost-wakeup race between `Wait` returning and `Reset()`.
2. **Presence probe enumerates everything** — `TransportConnectionWatchdog` polls at 1 Hz per transport (`:163`), and `SerialStreamTransport.cs:586-605` runs a **full `SerialPort.GetPortNames()` enumeration first** (allocating a `string[]` + a string per port), with the cheap `File.Exists(portName)` device-node check only as a fallback.
3. **A new OS thread per SCPI exchange** — every text exchange constructs a `StreamMessageConsumer<string>` (`TextExchangeEngine.cs:266`) whose `Start()` spawns a dedicated thread (`StreamMessageConsumer.cs:179`); a status-polling MCP or desktop session pays thread create + join on every command.

Each connected device pins 2 dedicated threads + a timer; a 10-device fleet idles at 20 threads, 100 producer wakeups/s, and 10 full port enumerations/s.

## Why
Long-running consumers (desktop, MCP server) hold devices connected for hours; this is measurable background CPU, allocation, and thread pressure on a system doing nothing.

## How
- Fix the producer signal ordering (or move to `SemaphoreSlim`/`Channel`) so the timeout can be removed; keep `Stop`/`StopSafely` latency bounds.
- Check `File.Exists` first on Unix device-node paths; share one `GetPortNames()` snapshot per process per poll tick across watchdogs.
- Reuse/pool the text-consumer reader instead of a fresh thread per exchange.

## Success criteria
- [ ] Zero producer wakeups while the queue is empty; `Stop`/`StopSafely` still return within their current bounds (existing tests).
- [ ] `SerialPort.GetPortNames()` called ≤1×/s process-wide regardless of device count.
- [ ] No thread creation per SCPI exchange; threads-per-connected-device documented and covered by a test or doc note.

tylerkron:
## Bench re-measurement: all three items are settled — recommend closing

[...]

PR #698, opened 2026-08-28
perf(benchmarks): add a BenchmarkDotNet harness so the perf numbers are reproducible

### LINKED ISSUE #640: perf: no benchmark harness — nine perf tickets were measured by hand and nothing guards the wins
**Type**: Performance / Infrastructure
**Priority**: Medium

## What
There is no benchmark project in the repo:

```
$ find . -iname "*benchmark*" -not -path "*/obj/*" -not -path "*/bin/*"
(no matches)
```

Yet performance is a recurring, heavily-worked theme. Nine tickets, every one of them stating a specific number:

| Issue | Claim | State |
|---|---|---|
| #491 | idle devices burn CPU — 10 producer wakeups/s per device, a new OS thread per SCPI exchange | **open** |
| #485 | every SCPI text exchange pays **~700 ms** of fixed polling/teardown overhead | **open** |
| #490 | steady-state decode allocates per frame what it could cache | closed |
| #489 | all three SD parsers materialize the whole file before yielding the first sample | closed |
| #488 | `discover_devices` runs WiFi and serial sequentially | closed |
| #487 | Windows runs one uncached WMI query per COM port per sweep | closed |
| #486 | the serial probe wastes **~1 s per port** on teardown | closed |
| #157 | serial device discovery is too slow | closed |
| #122 | optimize `ExecuteTextCommandAsync` timing for SD operations | closed |

Every one of those numbers was produced by hand, on someone's machine, once. None of them is reproducible today, and nothing detects when one regresses.

The evidence that regression is a live concern rather than a theoretical one is already in the tracker: **#531** — `Record_AfterTheFirstSamplePerChannel_AllocatesNothing` — is an allocation assertion written as a unit test, and it went flaky ("fails on a full net10 run, passes in isolation") precisely because a correctness-test harness is the wrong instrument for measuring allocation. That test is a benchmark wearing a `[Fact]` attribute.

## Why
The decode path is the library's hot loop and its performance is a product characteristic, not an implementation detail. `StreamMessageConsumer` reads from the wire on a dedicated thread, frames protobuf out of a `List<byte>`, and `StreamFrameDecoder` fans each frame out to per-channel `SampleReceived` handlers. At the sample rates this hardware runs, per-frame allocation is what determines whether a long acquisition survives without GC pauses showing up as jitter in `AcquisitionStatistics` — the very metric #502 added so users could see it.

Without a harness:
- The wins from #490, #489, #486 have no guard. The next refactor that reintroduces a per-frame allocation ships silently.
- The open tickets #485 and #491 have no baseline to prove they were fixed. "~700 ms of fixed overhead" needs a before-and-after number, and right now that number can only be produced ad hoc again.
- The CI matrix widening (#635) will make platform-specific performance differences visible for the first time, with nothing to measure them with.

## How
Add a `src/Daqifi.Core.Benchmarks` console project referencing BenchmarkDotNet, excluded from `dotnet test` and not packed. Target the paths the closed tickets already identified as hot:

1. **Frame decode** — `StreamFrameDecoder` over a captured `DaqifiOutMessage` payload, with `[MemoryDiagnoser]`. This is the #490 and #531 territory; the allocation assertion in #531 should move here and out of the correctness suite.
2. **Protobuf framing** — `StreamMessageConsumer`'s buffer accumulate/drain against a scripted stream, covering the partial-frame path.
3. **SD log parsing** — all three parsers (protobuf, CSV, JSON) over a representative file, measuring both throughput and time-to-first-sample, which is what #489 actually changed.
4. **Channel scaling** — `ChannelScaling.Apply` / `AnalogChannel.GetScaledValue`, per-sample and cheap to regress.

Do not wire this into the PR-gating CI. Benchmark numbers on shared GitHub runners are too noisy to gate on, and a flaky perf gate would land straight back in the #634/#632 flake pattern. Run it on demand (`workflow_dispatch`) and/or before a release, and check the results into the ticket that claims a win. If a guard is wanted later, BenchmarkDotNet's `[MemoryDiagnoser]` allocation counts are far more stable across runners than timings — allocation is the number worth gating on, not milliseconds.

## Success criteria
- [ ] `src/Daqifi.Core.Benchmarks` exists, runs via `dotnet run -c Release`, and is excluded from the test and pack paths.
- [ ] Benchmarks cover frame decode, protobuf framing, all three SD parsers, and channel scaling, all with `[MemoryDiagnoser]`.
- [ ] The allocation assertion behind #531 is expressed as a benchmark; the flaky `[Fact]` version is retired or reduced to a coarse smoke check.
- [ ] Baseline numbers for the current `main` recorded in the repo (a short `README.md` in the benchmark project) so #485 and #491 have a before to compare against.
- [ ] A `workflow_dispatch` job that can run the suite on demand without gating PRs.

tylerkron:
## What was wrong

This repo has nine performance tickets and every one of them states a number — "~700 ms of fixed
overhead per SCPI exchange", "~1 s per port wasted on teardown", "the decode path allocates per
frame". Each of those numbers was produced by hand, once, on whoever's machine was handy. None of
them can be reproduced today, none of the wins has a guard, and the two tickets still open (#485,
#491) have no *before* to prove a fix against. The one attempt at pinning a number in place —
#531's allocation assertion written as a `[Fact]` — went flaky precisely because a correctness
test is the wrong instrument for measuring allocation.

## How it was fixed

A `src/Daqifi.Core.Benchmarks` project running BenchmarkDotNet over the paths those tickets
identified: stream frame decode, protobuf framing, the consumer's read/append/drain loop, all
three SD-card log parsers (each measured twice — full drain and time to first sample), and the two
per-sample scaling conversions. Everything reports under `[MemoryDiagnoser]`, and the decode,
framing and consumer families report per *frame*, so a number can be read straight against a
sample rate. Baseline figures for `main` are in the project README.

It runs on demand only — a new `Benchmarks` workflow with `workflow_dispatch` — and deliberately
does not gate pull requests. Timings on shared GitHub runners are noisy enough that a perf gate
would fail green PRs regularly, which is the exact flake pattern (#634, #632) this repo has spent
real effort escaping. If a gate is ever wanted, allocation is the number to gate on, not
milliseconds.

Two things a reviewer might push back on, so they are stated plainly:

- **It found a bug, and I did not fix it here.** The protobuf SD parser hands back its first
  sample after 1.6 ms and 6.7 MB where CSV and JSON take about 2 µs. That is not the configuration
  pre-scan (dropping its message limit from 512 to 8 changed nothing) — it is `BufferSize`: the
  reader parses every message in a 64 KB read before yielding the first, and the configuration
  pass parses that same chunk again. Filed as **#697**. Fixing it in this PR would have mixed
  "build the instrument" with "act on what it read".
- **The `[Fact]` from #531 stays.** A later PR already stabilized it, and it is a perfectly good
  coarse pass/fail guard. What it was never good at is producing a *number*, so it now carries a
  comment pointing at this project for that. Deleting a green test to satisfy a checkbox seemed
  worse than leaving it and saying what each instrument is for.

The project is out of `dotnet test` (`IsTestProject=false`) and out of pack
(`IsPackable=false`), each guarded by a test, because both would fail quietly: without the first,
CI's `dotnet test Daqifi.Core.sln` starts asking a console project for tests; without the second,
`dotnet pack` starts producing a Daqifi.Core.Benchmarks package.

## Verification

- Full solution green on both frameworks: **4,096 passed / 3 skipped** on net9.0 and net10.0, plus
  217 Mcp tests each. `dotnet test Daqifi.Core.sln` walks past the new project without touching it,
  which is the behaviour the guard test pins.
- The suite itself runs clean end to end (15 benchmarks, ~4.5 min); every table in the README is
  from an actual run on this branch's code.
- No production code changed — the only edit outside the new project is a comment in
  `AcquisitionStatisticsTests` — so no bench validation applies.

closes #640

Not merging — this is for your review.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -53,11 +53,19 @@ jobs:
 
     # BenchmarkDotNet builds and launches its own process per benchmark, so this is
     # `dotnet run` rather than a build-then-test split.
+    #
+    # The filter arrives through the environment, not through ${{ }} interpolation into
+    # the script. `filter` is a free-text input, so interpolating it would splice whatever
+    # was typed straight into the shell — a value containing a quote closes the quoting and
+    # the rest runs as commands, on a runner holding this repo's token. `job` is a choice
+    # input with two fixed options, so it has no such freedom and is interpolated directly.
     - name: Run benchmarks
+      env:
+        FILTER: ${{ inputs.filter }}
       run: |
         dotnet run -c Release \
           --project src/Daqifi.Core.Benchmarks/Daqifi.Core.Benchmarks.csproj \
-          -- --filter '${{ inputs.filter }}' \
+          -- --filter "$FILTER" \
              ${{ inputs.job == 'short' && '--job short' || '' }} \
              --exporters github markdown json
 
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,81 @@
+name: Benchmarks
+
+# On demand only, never on a pull request (issue #640). Timings on a shared GitHub
+# runner are far too noisy to gate a merge on: a perf check that fails a green PR
+# once a week lands straight back in the #634/#632 flake pattern, and the fix for
+# that is not a wider tolerance, it is not gating. Run this before a release, or
+# either side of a change that claims a performance win, and paste the table into
+# the ticket that claims it.
+on:
+  workflow_dispatch:
+    inputs:
+      filter:
+        description: 'BenchmarkDotNet filter, e.g. *Decode* or *SdCard*'
+        required: false
+        default: '*'
+        type: string
+      job:
+        description: 'BenchmarkDotNet job. "short" for a quick look, "default" for numbers worth recording.'
+        required: false
+        default: 'default'
+        type: choice
+        options:
+          - default
+          - short
+
+jobs:
+  benchmark:
+    name: benchmark
+    # Pinned to one OS on purpose. A benchmark number is only meaningful against
+    # another number from the same machine shape, and comparing a run on a Windows
+    # runner with a baseline taken on a Linux one would produce a "regression" that
+    # is nothing but the runner. The cross-platform question CI answers is
+    # correctness (the three-OS test matrix); this answers a different one.
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v7
+
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v6
+      with:
+        dotnet-version: |
+          9.0.x
+          10.0.x
+
+    - name: Cache NuGet packages
+      uses: actions/cache@v4
+      with:
+        path: ~/.nuget/packages
+        key: nuget-${{ runner.os }}-${{ hashFiles('**/*.csproj', 'Directory.Build.props') }}
+        restore-keys: |
+          nuget-${{ runner.os }}-
+
+    # BenchmarkDotNet builds and launches its own process per benchmark, so this is
+    # `dotnet run` rather than a build-then-test split.
+    - name: Run benchmarks
+      run: |
+        dotnet run -c Release \
+          --project src/Daqifi.Core.Benchmarks/Daqifi.Core.Benchmarks.csproj \
+          -- --filter '${{ inputs.filter }}' \
+             ${{ inputs.job == 'short' && '--job short' || '' }} \
+             --exporters github markdown json
+
+    # The whole point of a manual run: the tables have to be readable afterwards
+    # without re-running anything.
+    - name: Summarize
+      if: always()
+      run: |
+        for report in BenchmarkDotNet.Artifacts/results/*-report-github.md; do
+          [ -e "$report" ] || continue
+          cat "$report" >> "$GITHUB_STEP_SUMMARY"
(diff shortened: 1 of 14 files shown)


tylerkron:

Head `54bef22`: 2 Qodo rounds. Round 1 raised two valid findings, both fixed in `54bef22`:

1. **Fixed timestamps corrupted iterations (Medium, correctness).** The decode benchmarks baked a timestamp into each frame, so every invocation replayed the same tick sequence into a decoder left at the end of the previous one — the first frame of each invocation took a backwards-step path the other 999 did not. Frames are now stamped from a monotonic per-case counter as they replay, so the device reads as one continuous 1 kHz acquisition, 32-bit rollover included. Sub-nanosecond cost in the measured loop, no allocation, and the measured numbers are unchanged (257.7 / 329.8 / 814.6 ns vs 257.9 / 332.2 / 820.5 before).
2. **Shell injection in the workflow (Medium, security).** The free-text `filter` input was interpolated into a single-quoted shell argument, so a filter containing a quote would have run as commands on a runner holding the repo token. It now arrives through the environment. The `job` input is a two-option choice, so it stays interpolated.

Round 2 on `54bef22` came back Bugs (0) with both findings marked resolved, 0 unresolved review threads, re-checked after settling. All four CI jobs green.

The README baseline was regenerated from this head, so the numbers in it are the ones this code produces. No production code changed in this PR, so no bench validation applies.

Not merging — this is for your review.

PR #699, opened 2026-08-28
perf(sdcard): stop the protobuf log parser decoding a whole 64 KB buffer before its first sample

### LINKED ISSUE #697: perf(sdcard): the protobuf log parser spends 1.6ms and 6.7MB before its first sample, while CSV and JSON take 2us

tylerkron:
## What was wrong

Opening an SD card recording and asking for the **first** sample was slow for `.bin` logs and fast
for everything else. On a 10,000-row, 8-channel recording the protobuf parser took **1.6 ms and
allocated 6.7 MB** before handing back one sample. The CSV and JSON parsers answered the same
question in about **2 µs**.

That gap matters for the interactive case — showing the first rows of a recording while the rest
streams in, which is what the lazy parsers were built for in #489. Laziness landed for CSV and JSON
but not for protobuf, and nobody noticed because "lazy" had been checked as "does not load the whole
file" rather than measured as a number.

## How it was fixed

Two independent causes, and neither is the configuration scan limit that the shape of the code
points at:

- The reader pulled a 64 KB buffer, decoded **every one of the ~1,300 frames in it**, and only then
  started handing them over one at a time. Now it decodes one frame at a time and passes each on as
  it appears.
- The configuration pre-scan always read 512 messages before deciding anything. When a log opens
  with a status message that states the timestamp clock — the common case — that message settles the
  configuration by itself and the scan's result is thrown away unread. It now stops as soon as
  nothing later in the file could change the answer.

**Behaviour is unchanged, and that is the part worth checking.** A status message stating the clock
already made the follow-on scan dead code — the `config == null || TimestampFrequency == 0` guard
never fires — so stopping there cannot alter any result. That is exactly the condition the scan
breaks on. Every existing test passed untouched.

The reviewable risk is elsewhere: `ProtobufMessageParser`'s resync, gap-gate and leading-noise rules
are subtle and were arrived at through several field failures (#268 among them). Rather than write a
second decode path beside them, `ParseMessages` is now a thin loop over the new stepper, so there is
still exactly one definition of where a frame begins. A theory test re-runs the whole corpus of
awkward buffers — echo-wrapped frames, null-byte runs, oversized prefixes, trailing partials —
through both entry points and demands identical frames and identical consumed counts.

The stepper is `internal`, so the public API surface is untouched and this does not disturb #680.

## Result

From `src/Daqifi.Core.Benchmarks`, same machine, default job:

| Method | Before | After | |
| --- | ---: | ---: | --- |
| `ProtobufTimeToFirstSample` | 1,565.6 µs | **9.4 µs** | 167× |
| `ProtobufTimeToFirstSample` allocated | 6,753 KB | **263 KB** | 26× |
| `ProtobufDrainAll` | 4,504.4 µs | **2,091.4 µs** | 2.2× |
| `ProtobufDrainAll` allocated | 16,880 KB | **13,170 KB** | 1.3× |

`CsvTimeToFirstSample` is 2.0 µs on the same run, so first-sample latency is now 4.7× CSV rather
than 800× — inside the order of magnitude the issue asked for. Draining the whole file got 2.2×
faster as a side effect, since the per-chunk list and message wrappers are gone.

## Verified

- Full suite green on net9.0 and net10.0: 4,116 passed in `Daqifi.Core.Tests`, 217 in
  `Daqifi.Mcp.Tests`, 0 failures. Release build is warning-clean.
- 17 new tests. The regression guard was checked against unmodified `main`, where it fails as
  intended: 6,144 bytes read to produce one sample, against ≤1,024 with the fix.
- No hardware needed — this is file parsing, so the benchmark is the measurement.

closes #697

### DIFF
--- a/src/Daqifi.Core.Tests/Communication/Consumers/ProtobufMessageParserSteppingTests.cs
+++ b/src/Daqifi.Core.Tests/Communication/Consumers/ProtobufMessageParserSteppingTests.cs
@@ -1,0 +1,197 @@
+using System.Text;
+using Daqifi.Core.Communication.Consumers;
+using Daqifi.Core.Communication.Messages;
+using Google.Protobuf;
+
+namespace Daqifi.Core.Tests.Communication.Consumers;
+
+/// <summary>
+/// Covers <see cref="ProtobufMessageParser.TryReadFrame"/>, the stepping entry point added for
+/// issue #697 so <c>SdCardFileParser</c> can take one frame at a time instead of paying for a
+/// whole 64 KB read buffer before its first sample.
+/// </summary>
+/// <remarks>
+/// Stepping and <see cref="ProtobufMessageParser.ParseMessages(byte[], out int)"/> run the same
+/// frame-boundary code, so the risk this file guards is that they stop doing so: the resync,
+/// gap-gate and leading-noise rules are subtle and were arrived at through four separate field
+/// failures (#268 among them). The equivalence theory below is therefore the important test —
+/// it re-runs the whole existing corpus of awkward buffers through the stepping API and demands
+/// the same frames and the same consumed count.
+/// </remarks>
+public class ProtobufMessageParserSteppingTests
+{
+    [Fact]
+    public void TryReadFrame_WithMultipleFrames_ReturnsOneFramePerStep()
+    {
+        // The property the SD card parser depends on: asking for a frame decodes one frame,
+        // not every frame in the buffer. Before #697 the only entry point parsed the lot.
+        var parser = new ProtobufMessageParser();
+        var data = Frames(7, 9, 11);
+
+        var offset = 0;
+        var timestamps = new List<uint>();
+
+        for (var i = 0; i < 3; i++)
+        {
+            var step = parser.TryReadFrame(data, offset, timestamps.Count > 0, out var frame, out var nextOffset);
+
+            Assert.Equal(ProtobufFrameStep.Frame, step);
+            Assert.NotNull(frame);
+            Assert.True(nextOffset > offset, "A decoded frame must advance the offset.");
+
+            timestamps.Add(frame!.MsgTimeStamp);
+            offset = nextOffset;
+        }
+
+        Assert.Equal(new uint[] { 7, 9, 11 }, timestamps);
+
+        // Exhausted, and the whole buffer was consumed.
+        var end = parser.TryReadFrame(data, offset, anyFrameRead: true, out var none, out var endOffset);
+        Assert.Equal(ProtobufFrameStep.EndOfBuffer, end);
+        Assert.Null(none);
+        Assert.Equal(offset, endOffset);
+        Assert.Equal(data.Length, offset);
+    }
+
+    [Fact]
+    public void TryReadFrame_StoppingEarly_DecodesNothingBeyondTheFrameItWasAskedFor()
+    {
+        // The stepper is only useful if abandoning it costs nothing further. A buffer whose
+        // tail is unparseable garbage still hands back the leading frame, because that tail is
+        // never looked at: had the parser decoded ahead, the garbage would be in its way.
+        var parser = new ProtobufMessageParser();
+        var data = Frames(42).Concat(Enumerable.Repeat((byte)0xFF, 4096)).ToArray();
+
+        var step = parser.TryReadFrame(data, 0, anyFrameRead: false, out var frame, out _);
+
+        Assert.Equal(ProtobufFrameStep.Frame, step);
+        Assert.Equal(42u, frame!.MsgTimeStamp);
+    }
+
+    [Fact]
+    public void TryReadFrame_WithTrailingPartialFrame_PreservesItInsteadOfAdvancing()
+    {
+        // A frame split across reads must be left for the caller to complete, so EndOfBuffer
+        // reports the offset unchanged — that is what makes nextOffset double as the consumed
+        // count for the caller's carry-over buffer.
+        var parser = new ProtobufMessageParser();
+        var whole = Frames(7, 9);
+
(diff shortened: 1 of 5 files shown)


tylerkron:
## Bench-tested on a real Nq1 (fw 3.7.2, USB/serial)

The PR body said "no hardware needed — this is file parsing." That was **understated**:
`ProtobufMessageParser` is also on the live streaming decode path via `StreamMessageConsumer`, so
this change touches live acquisition, not just SD log parsing. Bench-tested both paths, each time
against a CLI built from this branch **and** a second CLI built from `origin/main`, so every result
below is a differential rather than a "looks fine".

### SD log parsing — byte-identical output on real firmware data

Four genuine firmware-written `.bin` logs pulled off the card, parsed and exported to CSV by both
cores, then compared with `cmp`:

| Log | Size | Result |
| --- | ---: | --- |
| `log_20260826_165658.bin` | 9,784 B | identical, 467 rows |
| `log_20260803_192314.bin` | 12,691 B | identical, 508 rows |
| `log_20260812_165009.bin` | 3,706 B | identical, 310 rows |
| concatenated real frames | **68,224 B** | identical, 3,188 rows |

The 68 KB case is the one that matters: it exceeds the 64 KB read buffer, so it exercises the
multi-chunk carry-over — the straddling-frame handling this PR restructured — on real frame shapes
rather than synthetic ones. SD listing also verified (54 files).

### Live streaming — equivalent framing, no dropped or garbled frames

| | this branch | `origin/main` |
| --- | ---: | ---: |
| 100 Hz / 10 s, 3 ch — samples | 792 | 792 |
| tick delta | 420,000 ±23 | 420,000 ±20 |
| deltas implying a dropped frame | **0** | **0** |
| malformed rows | **0** | **0** |
| 1000 Hz / 10 s, 8 ch — samples | 7,930 | 7,929 |
| malformed rows | **0** | **0** |

At 100 Hz the ±23-tick spread is 0.55 µs of device sampling jitter at 42 MHz, not framing: a dropped
frame would show as a ~840,000 delta and there were none in either core.

### Two known firmware behaviours, reproduced identically on both cores

Neither is caused by this PR — both were confirmed against `origin/main` in the same session:

- **79% of requested rate** (792 samples for a 10 s 100 Hz run) — the fw 3.7.2 clock mismatch,
  daqifi-firmware#716.
- **~40% doubled intervals at 1000 Hz** (3,197 on this branch, 3,415 on `main`) — the high-rate
  duplicate-timestamp behaviour, daqifi-firmware#717. Slightly *fewer* on this branch, which is run
  to run variation, not an improvement claim.

Also hit the firmware#703 SD low-heap stall mid-session (downloads returning 0 and 19 bytes); it
reproduces on `main`, predates this work, and cleared with a device reboot.

PR #705, opened 2026-08-29
chore(api): check the packaged API against the release on nuget.org, not just against a file in the PR

### LINKED ISSUE #636: chore(api): nothing tracks the public API surface, but ADR 0002 makes source compatibility a promise
**Type**: Architecture / Release process
**Priority**: High

## What
`docs/adr/0002-binary-compatibility-policy.md` commits the library to a guarantee:

> Semver as practiced by this library (`README.md`, "For maintainers") therefore tracks **source compatibility** for its public API.

Nothing in the build enforces it. There is no `PublicAPI.Shipped.txt`, no `Microsoft.CodeAnalysis.PublicApiAnalyzers` reference, and no `<EnablePackageValidation>` in any of the four `.csproj` files:

```
$ grep -rn "PublicAPI\|PackageValidation\|ApiCompat" --include="*.csproj" .
(no matches)
```

The surface being promised is not small — `src/Daqifi.Core` declares **231 public types**, and the generated `Daqifi.Core.xml` carries 2,969 documented members (1,263 methods, 914 properties, 406 fields, 332 types, 54 events).

Today, removing a public method, narrowing a parameter type, or adding a member to a public interface produces a clean green build and a clean green CI run. The break surfaces when a consumer — `daqifi-desktop`, `Daqifi.Mcp`, or an external integrator — next restores the package.

## Why
This is not hypothetical for this repo; it is the standing blocker on live tickets:

- **#616** — "IDevice.IpAddress puts a network-only fact on the transport-agnostic device interface (**fix would break public API**)"
- **#612** — "Two device-abstraction leaks whose **fix would break public API**: dead `ProtobufMessageType.SdCard`, and `IsUsbConnection` as a transport type-test"
- **#557** — the ADR's own origin: an appended positional-record field, found by an adversarial audit on PR #554 rather than by tooling.

Both open tickets are parked specifically because the API break is understood to be costly, yet there is no artifact that says what the current surface *is*. A reviewer assessing "does this PR break the API?" has to reason it out from the diff every time — which is how #557 got to the point of needing an ADR.

A checked-in API surface file inverts that: a break becomes a **required, reviewable diff to `PublicAPI.Shipped.txt`** in the same PR. It makes the ADR's policy mechanically visible instead of a norm someone has to remember, and it turns #612/#616 from "we think this breaks something" into "here is the exact list of what breaks."

## How
Two options; they compose, and option 1 is the higher-value half.

1. **`Microsoft.CodeAnalysis.PublicApiAnalyzers`** on `Daqifi.Core`. Adds `PublicAPI.Shipped.txt` / `PublicAPI.Unshipped.txt`; RS0016 fails the build on any undeclared public member. With `TreatWarningsAsErrors` already `true` this needs no extra wiring. Seed `Shipped.txt` from the current release tag so the baseline is what actually shipped, and let `Unshipped.txt` accumulate between releases — it doubles as the release-notes checklist the ADR asks for ("It is called out in the release notes for the version that ships it").

2. **`<EnablePackageValidation>true</EnablePackageValidation>`** with `<PackageValidationBaselineVersion>` set to the last published NuGet version. This catches the *binary* shape too, including the positional-record arity case from #557 — which is exactly the failure the ADR documents as out-of-guarantee but wants surfaced in release notes.

Worth noting the ADR explicitly declines compatibility *shims*; neither option adds one. They only make the break visible at build time.

## Success criteria
- [ ] `PublicAPI.Shipped.txt` checked in for `Daqifi.Core`, seeded from the most recent published version.
- [ ] Adding, removing, or changing a public member fails the build until the API file is updated in the same PR.
- [ ] `EnablePackageValidation` with a baseline version, so the #557 record-arity shape is reported rather than discovered by audit.
- [ ] `CONTRIBUTING.md` gains a short note on updating the API files, next to the existing `ConfigureAwait(false)` section.
- [ ] #612 and #616 can each cite the exact surface diff their fix requires.

tylerkron:
## What was wrong

ADR 0002 promises source compatibility for `Daqifi.Core`'s public API. PR #680 made that promise visible: the surface is checked in as `PublicAPI.Shipped.txt`, and RS0016/RS0017 fail the build whenever the code and that file disagree.

But both files travel in the PR. Delete a public member *and* its line in `PublicAPI.Shipped.txt` in the same change and the build is green — which is precisely the shape of a break ADR 0002 is meant to catch. A deliberate break and an accidental one look identical to the build, and the accidental one reaches a consumer (`daqifi-desktop`, `Daqifi.Mcp`, an external integrator) on their next restore.

I confirmed this on main before changing anything: narrowing `DeviceMetadata.IpAddress`'s setter to `internal` and deleting its one line from `PublicAPI.Shipped.txt` builds clean, 0 warnings, 0 errors.

## How it's fixed

Compare against something the PR cannot edit: the package already on nuget.org.

`Daqifi.Core` now sets `EnablePackageValidation` with `PackageValidationBaselineVersion` pinned to **1.7.0**, the last published version, and CI packs the library so ApiCompat compares the packaged assemblies — both `net9.0` and `net10.0` — against that download. The same removal above now fails with `CP0002` on both frameworks.

That leaves the source analyzer as the fast, precise, everyday check and package validation as the backstop that answers "does this still work for someone who already depends on 1.7.0?"

A second CI step keeps the baseline honest: it resolves the newest stable version from the nuget.org index and fails when the pinned baseline is behind it. Without that, the check quietly narrows over time — ApiCompat can only report a member the baseline package actually contains, so everything added since the pinned version would fall outside the comparison entirely.

## Things a reviewer may want to push back on

**`CP0003` is suppressed.** It compares assembly *versions*, and this repo carries none — the version is stamped only by `release.yml` from the git tag. Every other build ships the SDK's `1.0.0` placeholder, so `CP0003` reports "lower than the 1.7.0 baseline" on every build, including ones that change no API at all. Suppressing it keeps the rules that carry signal here (`CP0002`, `CP1002`) readable. Version ordering is decided upstream by which tag the release is cut from.

**The gate is in `ci.yml`, not `release.yml`.** Package validation only runs inside `dotnet pack`, and `--no-build` skips the ApiCompat targets entirely — verified: with `--no-build` the pack is silent on a break that otherwise produces `CP0002`. `release.yml` packs `--no-build` on purpose, so its assemblies keep the version stamped by the earlier build step; removing that flag would republish the assemblies as `1.0.0.0`. So the PR/main gate is the right place, and everything reaching a release tag has passed it. `PublicApiTrackingTests` asserts the CI pack step never grows a `--no-build`.

**The baseline-freshness check is a failure, not a warning.** After a release is published, CI goes red on every PR until the baseline bump lands. That is the intent — the bump PR itself is green, and the alternative is a check that silently covers less over time. The bump belongs in the same change that moves `Unshipped` entries into `Shipped`.

**Prereleases are excluded when deciding what is "newest".** `release.yml` accepts prerelease tags and the index lists them, but a prerelease is not what a consumer restores by default, so it must not become the version every other PR is required to match.

**It costs one extra compile.** `dotnet pack` defaults to Release, so the step rebuilds `Daqifi.Core` rather than reusing the Debug output. One project, on the ubuntu leg only — the packaged shape is identical on all three. Measured at ~6s in CI.

## Verification

- Mutation proof for package validation: setter narrowed to `internal` **and** its `PublicAPI.Shipped.txt` line deleted → `dotnet build` green, `dotnet pack` red with `CP0002` on `net9.0` and `net10.0`.
- The pre-existing analyzer gate re-verified in both directions: a new public type → `RS0016`; a narrowed setter → `RS0017`.
- Baseline-freshness check: pinning 1.6.0 against the published 1.7.0 exits 1 with the `::error::` annotation. Prerelease filter checked against a simulated index and the live one.
- Every new test checked to fail when its wiring is removed.
- Full suite green on net9.0 + net10.0: 4152 + 217 per framework, 0 failed. CI green on all three OS legs, with both new steps passing.

No bench run — this is a build-configuration change with no device path.

Closes #636

Not merging — for review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,14 +85,26 @@ jobs:
     # Deliberately a failure, not a warning: after a release is published this
     # goes red until the bump lands, which is the point. It costs one unauthenticated
     # HTTP request to the index the pack step already downloads from.
+    #
+    # Stable versions only. release.yml accepts prerelease tags (`1.8.0-beta.1`),
+    # and the flat-container index lists them alongside the rest - but a prerelease
+    # is not what a consumer restores by default, so it must not become the
+    # baseline every other PR is then required to match. Filtering them here is
+    # also what keeps the baseline parseable as a plain three-part version, which
+    # PublicApiTrackingTests asserts.
     - name: Check the package validation baseline is the latest published version
       if: matrix.os == 'ubuntu-latest'
       shell: bash
       run: |
         baseline=$(dotnet msbuild src/Daqifi.Core/Daqifi.Core.csproj \
           -getProperty:PackageValidationBaselineVersion -p:TargetFramework=net10.0 | tr -d '[:space:]')
-        latest=$(curl -sSf https://api.nuget.org/v3-flatcontainer/daqifi.core/index.json | jq -r '.versions[-1]')
+        latest=$(curl -sSf https://api.nuget.org/v3-flatcontainer/daqifi.core/index.json \
+          | jq -r '[.versions[] | select(contains("-") | not)] | last')
         echo "baseline=$baseline latest=$latest"
+        if [ -z "$latest" ] || [ "$latest" = "null" ]; then
+          echo "::error::Could not resolve a stable published version of Daqifi.Core from nuget.org."
+          exit 1
+        fi
         if [ "$baseline" != "$latest" ]; then
           echo "::error::PackageValidationBaselineVersion is $baseline but Daqifi.Core $latest is published." \
                "Everything added since $baseline is outside the ApiCompat comparison." \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,31 @@ jobs:
     - name: Test
       run: dotnet test --no-build --no-restore Daqifi.Core.sln
 
+    # The baseline is only as good as it is current. ApiCompat can only report a
+    # member the baseline package actually contains, so every API added since the
+    # pinned version sits outside the comparison - an out-of-date baseline is a
+    # narrower check, not a stricter one, and the step below would keep passing
+    # while covering less and less. Bumping it is part of cutting a release
+    # (CONTRIBUTING.md); this is what makes forgetting visible instead of silent.
+    #
+    # Deliberately a failure, not a warning: after a release is published this
+    # goes red until the bump lands, which is the point. It costs one unauthenticated
+    # HTTP request to the index the pack step already downloads from.
+    - name: Check the package validation baseline is the latest published version
+      if: matrix.os == 'ubuntu-latest'
+      shell: bash
+      run: |
+        baseline=$(dotnet msbuild src/Daqifi.Core/Daqifi.Core.csproj \
+          -getProperty:PackageValidationBaselineVersion -p:TargetFramework=net10.0 | tr -d '[:space:]')
+        latest=$(curl -sSf https://api.nuget.org/v3-flatcontainer/daqifi.core/index.json | jq -r '.versions[-1]')
+        echo "baseline=$baseline latest=$latest"
+        if [ "$baseline" != "$latest" ]; then
+          echo "::error::PackageValidationBaselineVersion is $baseline but Daqifi.Core $latest is published." \
+               "Everything added since $baseline is outside the ApiCompat comparison." \
+               "Bump it in src/Daqifi.Core/Daqifi.Core.csproj."
+          exit 1
+        fi
+
     # The public API surface is guarded twice (issue #636). RS0016/RS0017 guard it
     # at compile time against the checked-in PublicAPI.*.txt, which the Build step
     # above already covers - but those files are part of the PR, so a PR that
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,29 @@ jobs:
     - name: Test
       run: dotnet test --no-build --no-restore Daqifi.Core.sln
 
+    # The public API surface is guarded twice (issue #636). RS0016/RS0017 guard it
+    # at compile time against the checked-in PublicAPI.*.txt, which the Build step
+    # above already covers - but those files are part of the PR, so a PR that
+    # deletes an entry and the member together compiles clean. This step is the
+    # second half: it compares the packaged assemblies against what Daqifi.Core
+    # actually published to nuget.org (the baseline version in Daqifi.Core.csproj),
+    # which a PR cannot edit, and fails on a member the baseline had.
+    #
+    # `--no-build` is deliberately absent: it skips the ApiCompat targets outright,
+    # so a pack that uses it validates nothing. That is also why release.yml's pack
+    # is not the gate - it packs `--no-build` so the assemblies keep the version
+    # stamped from the git tag. Everything reaching a release tag has passed this
+    # step first, on the PR and again on main.
+    #
+    # Ubuntu only: the packaged shape is identical on all three legs, so the other
+    # two would pay to compute the same answer. `dotnet pack` defaults to Release,
+    # so this does compile Daqifi.Core once more rather than reusing the Debug
+    # output above - one project, a few seconds, on one leg. The package itself is
+    # thrown away; only the validation result matters.
+    - name: Validate packaged API against the last published release
+      if: matrix.os == 'ubuntu-latest'
+      run: dotnet pack src/Daqifi.Core/Daqifi.Core.csproj --no-restore --output "${{ runner.temp }}/api-validation"
+
     # Daqifi.Core.Tests.csproj collects coverage on every run, so the cobertura
     # report already exists by this point; the next two steps are only about not
     # throwing it away when the job ends. Reporting only — deliberately no
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -75,6 +75,30 @@ name in the message is already the exact line to add - those types sit in the gl
 And a *removal* is never automatic: deleting the entry is the deliberate act of declaring a
 breaking change, and ADR 0002 says what that costs.
 
+### The published package is the second opinion
+
+`PublicAPI.*.txt` lives in the repo, so a PR that removes a public member *and* its entry
+compiles clean - the deliberate act above looks identical to an accidental one. So
+`Daqifi.Core` also sets `EnablePackageValidation` with `PackageValidationBaselineVersion`
+pinned to the last version on nuget.org. CI packs the project (`Validate packaged API against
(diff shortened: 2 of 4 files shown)


tylerkron:
Round 1 finding accepted — it was a genuine logic error, not just a wording problem.

I had claimed in two places (the `Daqifi.Core.csproj` comment and `CONTRIBUTING.md`) that leaving `PackageValidationBaselineVersion` behind "only makes the check stricter." That is wrong for exactly the reason given: ApiCompat can only report a member the baseline package actually contains, so an old baseline is a *narrower* check. Every API added after the pinned version is outside the comparison, and could be removed together with its `PublicAPI.Shipped.txt` entry with both guards still green — reopening the same loophole this PR exists to close.

Fixed in 526b554:

- Corrected the claim in the csproj comment and in `CONTRIBUTING.md`; the bump is now stated as required, with the narrower-not-stricter reasoning spelled out.
- Added a CI step that resolves the newest published version from the nuget.org flat-container index and fails when the pinned baseline is behind it. Nothing in the repo records which version is current, so the question has to be asked of nuget.org — and the pack step already downloads from that same index, so it is not a new dependency.
- The test keeps the parseability assertion (that the baseline is pinned at all) and gained `Ci_FailsWhenTheBaselineHasFallenBehindTheLatestRelease`, which asserts CI still asks that question.

Verified by pinning the baseline to 1.6.0 against the published 1.7.0: the step exits 1 with the `::error::` annotation naming the drift. Restored to 1.7.0, which passes.

One deliberate consequence worth flagging: this is a failure, not a warning, so CI goes red on every PR between a release being published and the baseline bump landing. That is the intent — the bump PR itself is green, and the alternative is a check that quietly covers less over time.

Full suite green on net9.0 + net10.0 after the change: 4152 + 217 per framework, 0 failed.

tylerkron:
Round 2 — both findings accepted, fixed in 5d31439.

**1. Prerelease baseline cannot pass.** Correct, and it was worse than a mismatch: the two halves of this change would have deadlocked against each other. `release.yml` accepts prerelease tags, the flat-container index lists them alongside stable versions, and my freshness check took the last entry verbatim — so publishing `1.8.0-beta.1` would have demanded a baseline that `Version.TryParse` rejects. Neither leaving the baseline alone nor bumping it correctly could have passed CI.

The query now filters prereleases out (`select(contains("-") | not)`), so the baseline is always the newest **stable** release. That is the right answer independently of the parsing problem: a prerelease is not what a consumer restores by default, so it should not become the version every other PR is required to match. Filtering is also what keeps the baseline parseable, so `Version.TryParse` is now deliberate rather than accidental — the test comment says so. I also added a guard for an empty/null result rather than silently comparing against nothing.

Verified against a simulated index `["1.6.0", "1.7.0", "1.8.0-beta.1"]` → `1.7.0`, and against the live index → `1.7.0`.

**2. CI guard tests only strings.** Also correct — asserting that two strings appear *somewhere in the file* would have stayed green with the comparison or the `exit 1` deleted, which is the exact silent disablement the test exists to catch.

It now locates the step by name, extracts its own `run:` block, and asserts each part the enforcement actually rests on: the property lookup, the index URL, the prerelease filter, the `"$baseline" != "$latest"` comparison, and `exit 1`. Verified by replacing both `exit 1` lines with an `echo` — the test fails.

I stopped short of executing the shell from the test. That would mean spawning bash on all three matrix legs for a build-configuration guard, and this file is explicit that its reach is deliberately shallow rather than a second, worse copy of the thing it guards. Scoping the assertions to the step and to the load-bearing tokens closes the gap you identified without that cost.

Full suite green on net9.0 + net10.0: 4152 + 217 per framework, 0 failed. CI was green on 526b554 including the new pack and baseline steps (the pack step runs in ~6s).

tylerkron:

Verified against head `5d31439`: Qodo summary reads Bugs (0) / Rule violations (0) / Skill insights (0) with all three findings struck through as resolved, and all 3 `qodo-code-review` review threads are resolved (0 unresolved). Re-checked after a settle interval — both surfaces unchanged. CI run 33258223069 is green on that exact SHA across all three OS legs, with both new gates (`Check the package validation baseline` and `Validate packaged API against the last published release`) running and passing on the ubuntu leg.

Re-validated against current `main`: the branch already contains #703 (`d75d82a`, build-time IDE0161 + CA1707), so `git merge origin/main` is a no-op and the green run above already exercised the merged tree. Confirmed locally that `dotnet pack src/Daqifi.Core/Daqifi.Core.csproj` (without `--no-build`, so ApiCompat actually runs) builds and packs both net9.0 and net10.0 clean under the new analyzer rules — the new style enforcement does not trip package validation. The `--no-build` trap holds as documented: `ci.yml` packs without it so validation runs, while `release.yml` packs with it to keep the tag-stamped version, which is why the gate lives in CI.

tylerkron:
**Negative control: the gate actually fails.** Verified independently on `5d31439` (macOS, SDK 10.0.203) — everything prior was positive evidence only (the tree packs clean), which looks identical whether the gate works or does nothing.

**Mutation** — the exact scenario #636 describes: deleted the shipped public parameterless ctor `TransportNotConnectedException()` *and* removed its `PublicAPI.Shipped.txt` entry, so RS0016/RS0017 stay green and the in-repo file check cannot be what catches it. (Chosen because it has zero callers inside `Daqifi.Core`, so the deletion compiles clean — the build succeeded and the `.nupkg` was even produced; only validation failed.)

**Result** — `dotnet pack` exits 1, on **both** target frameworks:

```
error CP0002: Member 'Daqifi.Core.Communication.Transport.TransportNotConnectedException.TransportNotConnectedException()'
  exists on [Baseline] lib/net10.0/Daqifi.Core.dll but not on lib/net10.0/Daqifi.Core.dll
error CP0002: ... same on lib/net9.0/Daqifi.Core.dll
error : API breaking changes found. If those are intentional, the APICompat suppression file can be
  updated by rebuilding with '/p:ApiCompatGenerateSuppressionFile=true'
```

Compared against the real published `daqifi.core.1.7.0.nupkg`, confirming the baseline resolves to the nuget.org artifact rather than anything in the PR. Full cycle: packs clean → mutated fails → **mutation reverted**, tree clean at `5d31439`, packs clean again.

---

*One discrepancy, non-blocking.* The comment on `Ci_PacksTheLibrarySoPackageValidationActuallyRuns` says `--no-build` makes the pack "silent". On SDK 10.0.203 it does **not** — `dotnet pack --no-build` against the mutated tree still reported both `CP0002`s and still exited 1. So validation is not skipped there, at least on this SDK. This doesn't weaken the PR: omitting `--no-build` in `ci.yml` is correct and strictly safer either way, and the assertion the test makes is still the right one. Only the stated rationale is narrower than claimed — likely SDK-version-dependent, so it may hold on whatever SDK it was originally observed on. Worth softening that comment if the file is touched again; not worth a push on its own.

